PR #123, opened 2015-02-07
X11: use the hwloc.m4-determined X11 flags properly

jsquyres:
Set new HWLOC_X11_CPPFLAGS macro with X_CFLAGS, and set HWLOC_X11_LIBS to all 3 of X_PRE_LIBS, X_LIBS, and X_EXTRA_LIBS.

Then be sure to use both of these macros when compiling lstopo.

This patch based on an email sent to me from @bgoglin with initial bug report from @scottatchley.

Specifically, I think Scott's initial patch was in the right direction, but I'm not sure it went far enough.  I don't have problems compiling on my Yosemite (10.10.2) OS X system, but:

1. I don't seem to have a sym link from /opt/X11 to /usr/X11, and
1. My CPP and linker seem to find the X11 headers/libraries in their default search paths.

Specifically, in my lstopo/Makefile:

```make
X_CFLAGS = 
X_EXTRA_LIBS = 
X_LIBS = 
X_PRE_LIBS = 
```

Hence, I wonder if the real solution is twofold:

1. Ensure that HWLOC_X11_LIBS is actually complete, and ensure that HWLOC_X11_CPPFLAGS exists.
2. Use both of those values properly in lstopo/Makefile.am.

@scottatchley Does this patch work for you?

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -9,7 +9,7 @@ dnl Copyright © 2004-2012 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright © 2004-2008 High Performance Computing Center Stuttgart,
 dnl                         University of Stuttgart.  All rights reserved.
-dnl Copyright © 2006-2014 Cisco Systems, Inc.  All rights reserved.
+dnl Copyright © 2006-2015 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright © 2012  Blue Brain Project, BBP/EPFL. All rights reserved.
 dnl Copyright © 2012       Oracle and/or its affiliates.  All rights reserved.
 dnl See COPYING in top-level directory.
@@ -876,7 +876,9 @@ EOF])
              AC_CHECK_HEADERS([X11/Xutil.h],
                 [AC_CHECK_HEADERS([X11/keysym.h],
                     [AC_DEFINE([HWLOC_HAVE_X11_KEYSYM], [1], [Define to 1 if X11 headers including Xutil.h and keysym.h are available.])])
-                     AC_SUBST([HWLOC_X11_LIBS], ["-lX11"])
+                     AS_IF([test "x$X_LIBS" = "x"], [X_LIBS="-lX11"])
+                     AC_SUBST([HWLOC_X11_CPPFLAGS], [$(X_CFLAGS)])
+                     AC_SUBST([HWLOC_X11_LIBS], ["$(X_PRE_LIBS) $X_LIBS $(X_EXTRA_LIBS) "])
                 ], [], [#include <X11/Xlib.h>])
             ])
          ])
--- a/utils/lstopo/Makefile.am
+++ b/utils/lstopo/Makefile.am
@@ -1,6 +1,6 @@
 # Copyright © 2009-2015 Inria.  All rights reserved.
 # Copyright © 2009-2012, 2014 Université Bordeaux
-# Copyright © 2009-2014 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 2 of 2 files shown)


config/hwloc.m4
```diff
@@ -873,7 +873,8 @@ EOF])
              AC_CHECK_HEADERS([X11/Xutil.h],
                 [AC_CHECK_HEADERS([X11/keysym.h],
                     [AC_DEFINE([HWLOC_HAVE_X11_KEYSYM], [1], [Define to 1 if X11 headers including Xutil.h and keysym.h are available.])])
-                     AC_SUBST([HWLOC_X11_LIBS], ["-lX11"])
+                     AC_SUBST([HWLOC_X11_CPPFLAGS], ["$(X_CFLAGS)"])
+                     AC_SUBST([HWLOC_X11_LIBS], ["$(X_PRE_LIBS) $(X_LIBS) $(X_EXTRA_LIBS) "])
```
jsquyres: I'm not actually sure that the quotes are needed for HWLOC_X11_CPPFLAGS and HWLOC_X11_LIBS -- I kept them because they were in the original line 876.

But I can't check for sure because my system has X11 in default paths, so this code path isn't executed.  

@scottatchley Can you check to see what happens on your machine?

scottatchley:
Hi Jeff,

I just found this. The patch does not work for me. It does not include -L$(X11)/lib or -lX11 when trying to build lstopo.

Scott

bgoglin:
Just got another similar report on Yosemite.
From what I see in the doc of AC_PATH_XTRA, -lX11 is not in any of X_LIBS/X_PRE_LIBS/X_EXTRA_LIBS, you still need to add it to LIBS. That would explain why @scottatchley didn't have -lX11 when linking lstopo. Not sure about the missing -L$(X11)/lib

scottatchley:
I am still using Mavericks. Would you like me to try a new tarball or from git?

jsquyres:
Ok, I just pushed a new commit onto this PR.  It works for me now on OS X Yosemitie (10.10.5) when I configure like this:

```sh
./configure --prefix=$bogus CPPFLAGS=-I/opt/local/include LDFLAGS=-L/opt/local/lib
```

(since my X11 libs are in /opt/local via MacPorts)

Does it work for you?

scottatchley:
On Sep 11, 2015, at 9:07 AM, Jeff Squyres <notifications@github.com> wrote:

> Ok, I just pushed a new commit onto this PR. It works for me now on OS X Yosemitie (10.10.5) when I configure like this:
> 
> ./configure --prefix=$bogus CPPFLAGS=-I/opt/local/include LDFLAGS=-L/opt/local/lib
> (since my X11 libs are in /opt/local via MacPorts)
> 
> Does it work for you?

Are you configuring hwloc only or ompi? (trying to avoid configuring ompi ;-))

jsquyres:
This is a pull request on just the hwloc repo (in the "open-mpi" org, but still just plain hwloc).  :smile:

bgoglin:
I still don't see -lX11 in your new patch?
Anyway, I sent another patch to Scott and my other bug reported earlier today, seems to work, so I am closing this one.

scottatchley:
For completeness, it works as long as the user has pkg-config installed via fink, macports, etc.

Having just XQuartz is not enough.

bgoglin:
I'll see if we can support cairo without pkg-config

jsquyres:
@bgoglin You mean you didn't see https://github.com/open-mpi/hwloc/pull/123/files#diff-92968398e4fd87bc82b2b1d08f51fc20R879 ?  :smile: 

Doesn't matter -- if the problem is fixed a different way, so be it.

PR #149, opened 2016-02-24
RFC: Remove libnuma dependency

bgoglin:
@jsquyres This came up during the OMPI dev meeting when discussing with @markalle. Platform-MPI currently dlopens libnuma.so instead of keeping the hardwired dependency in libhwloc.so. I figured we could try to just remove the libnuma dependency entirely. It's easy, except when redefining the missing syscalls. I still need to check whether syscall() is well supported with 6 parameters on reasonably old distros running on NUMA machines.

Likely for hwloc v2.0 only, except if somebody is sure this can safely go to v1.11.3.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -111,6 +111,7 @@ Version 2.0.0
     anymore, they should be passed in HWLOC_XMLFILE or HWLOC_SYNTHETIC instead.
   + HWLOC_COMPONENTS takes precedence over other environment variables
     for selecting components.
+  + Remove the dependency on libnuma on Linux.
 
 
 Version 1.11.2
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -572,13 +572,13 @@ EOF])
          AC_MSG_RESULT([yes])],
         [AC_MSG_RESULT([no])])
 
-    AC_MSG_CHECKING([for working syscall])
+    AC_MSG_CHECKING([for working syscall with 6 parameters])
     AC_LINK_IFELSE([
       AC_LANG_PROGRAM([[
           #include <unistd.h>
           #include <sys/syscall.h>
-          ]], [[syscall(1, 2, 3);]])],
-        [AC_DEFINE([HWLOC_HAVE_SYSCALL], [1], [Define to 1 if function `syscall' is available])
+          ]], [[syscall(0, 1, 2, 3, 4, 5, 6);]])],
+        [AC_DEFINE([HWLOC_HAVE_SYSCALL], [1], [Define to 1 if function `syscall' is available with 6 parameters])
          AC_MSG_RESULT([yes])],
         [AC_MSG_RESULT([no])])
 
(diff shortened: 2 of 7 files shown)


markalle:
Just for informational purposes, here's what we've been doing in our Platform-MPI build of hwloc:

We wanted to soften the runtime dependency on libnuma.so, so we added three functions as below, where "set_mempolicy_fptr" etc are function pointers that we try to dlopen()/dlsym() out of libnuma.so.  So if we are able to get those symbols, we just pass through and use the real function out of libnuma.so, but if the system didn't have libnuma.so I tried to pick return values that wouldn't break anything for a no-op.

This gave us a single build of libhwloc.so that would support the various memory binding calls if libnuma was availalble, but still function with slightly reduced functionality on systems without libnuma. (I'm not proposing this as the solution to use here though, it's just what we did.)

long
set_mempolicy(int a, unsigned long *b, unsigned long c)
{
  if (!numastub_is_init) { numastub_load_symbols(); }
  if (numastub_has_numa) {
    return set_mempolicy_fptr(a, b, c);
  }
  return 0;
}
// from browsing hwloc source the best thing we can do to avoid
// crashing their code is probably to return 0 and set *a to
// MPOL_DEFAULT which is 0.
long
get_mempolicy(int *a, unsigned long *b, unsigned long c, void *d, int e)
{
  if (!numastub_is_init) { numastub_load_symbols(); }
  if (numastub_has_numa) {
    return get_mempolicy_fptr(a, b, c, d, e);
  }
  *a = 0;
  return 0;
}
long
mbind(void *a, unsigned long b, int c, unsigned long *d, unsigned long e, unsigned f)
{
  if (!numastub_is_init) { numastub_load_symbols(); }
  if (numastub_has_numa) {
    return mbind_fptr(a, b, c, d, e, f);
  }
  return 0;
}

jsquyres:
@markalle Github pro tip: Put a "c" (with no quotes) after the triple-single-tick to start the verbatim to get C syntax highlighting.

jsquyres:
@bgoglin @markalle Interesting approaches.  I agree that removing the dependency of having libnuma-devel (or whatever the OS distro's package name for it is) will be a Good Thing; we have run across users who didn't have this installed and then complained when Open MPI's/hwloc's functionality is reduced.

I guess there's pros/cons to each approach:

1. @bgoglin's approach is more robust -- it doesn't even depend on `libnuma` being installed at all.  But it might be a little more fragile with older kernels.  It also hard-codes a bunch of values in hwloc (that are unlikely to ever change, but still...).
1. @markalle's approach is definitely more simplistic / less code / probably easier to maintain over time.  It does rely on having having `libnuma` installed (which may not be a big deal).

bgoglin:
Hard-coding values isn't a problem. Linus won't let the kernel guys change them since it would break the syscall ABI.

We dug into the glibc history and found that the only problem with the syscall macro with 6 arguments was a x86_64-specific problem that was fixed in glibc 2.3.4 in 2004. libnuma enabled it for glibc 2.11+ only because it was the glibc that was widely used when they made the change, but they didn't check earlier releases.

So everything actually looks good (although I am not 100% sure yet that backporting this to v1.11.x is acceptable).

bgoglin:
All my testing went well, so I merged this in master.

If somebody wants a backport in v1.x, please check on more "old" Linux machines (will glibc between 2.3.4 and 2.11, not only on x86_64) that tests/hwloc/hwloc_bind reports the same with and without commit e6a53bb.

PR #152, opened 2016-03-08
[RFC] Insert right objects before the left ones

bgoglin:
Not really useful as long as we don't have hundreds of children within a single object. The quadratic complexity is negligible on current machines since we usually have tens of children max. Mostly useful for large synthetic topologies so far.

Check whether the first commit in the PR decreases performance before merging.

Otherwise just keep that code here until machines really need that sort of optimization. The Linux backend will need a reverse foreach loop.


bgoglin:
This deprecated branch is saved in bgoglin/zz-right-to-left-insert in case somebody ever cares.

PR #191, opened 2016-06-26
[RFC] stop distributing some tests

bgoglin:
Our release tarballs are currently about 4MB, with about 700kB of tests/hwloc/linux tarballs. Not too bad. But things are expected to get worse with KNL tarballs for different MCDRAM/Clustering configs (100kB each), newer x86 CPUID tarballs, netloc stuff, etc.

This PR makes tests/hwloc/{linux,xml,x86}/ and utils/hwloc/test-hwloc-dump-hwdata/ support running optional tests under an "extra" subdirectory. Those tests would be in GIT but not in tarballs.

The same code could be used to move tests to another repository if git clones become too heavy (currently 35MB). Something like cloning hwloc-extra/ at the root the main hwloc clone and having tests subdirectories tests/hwloc/linux run additional tests in hwloc-extra/tests/hwloc/linux. We would warn at the end of configure when building from GIT without that extra repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,6 +116,7 @@ test-suite.log
 /tests/hwloc/nvml
 
 /tests/hwloc/linux/test-topology.sh
+/tests/hwloc/linux/extra/
 
 /tests/hwloc/linux/gather/test-gather-topology.sh
 
--- a/.gitignore
+++ b/.gitignore
@@ -124,6 +124,7 @@ test-suite.log
 /tests/hwloc/ports/liblstopo-port-windows.a
 
 /tests/hwloc/x86/test-topology.sh
+/tests/hwloc/x86/extra/
 
 /tests/hwloc/xml/test-topology.sh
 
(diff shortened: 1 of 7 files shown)


jsquyres:
Does the size of the tarball and/or git clone really matter?

Random data point: out of the ~7,500 MP3s in my music collection, the average size is 4.2MB.  The Open MPI web site reports that the hwloc 1.11.3 bz2 tarball is 4.11MB.  Meaning: hwloc is still smaller than an average MP3 file.

bgoglin:
Not sure about the git clone size.

Yes, the tarball size matters. There are people with slow internet access. And there are people who care about wasting power for downloading useless tests that they won't run :)

The current size is OK. But there are 15 possible KNL configs, which would add 180kB each. Testing all of them before a release is good. But that testing all of them on many different machines isn't very useful. The issue will get bigger as we add newer bigger sysfs tarballs (their size seems to increase between in n*log(n) with n=number of cores). And much bigger for netloc dumps of IB fabrics.

PR #192, opened 2016-06-26
plugins.h: fix "might be uninitialized" compiler warning

jsquyres:
This commit fixes a warning from clang/OS X:

```
  CC       topology.lo
In file included from /Users/jsquyres/git/hwloc/include/private/components.h:19:0,
                 from /Users/jsquyres/git/hwloc/include/private/private.h:28,
                 from topology.c:29:
topology.c: In function ‘hwloc_topology_load’:
/Users/jsquyres/git/hwloc/include/hwloc/plugins.h:439:6: warning: ‘filter’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   if (filter == HWLOC_TYPE_FILTER_KEEP_IMPORTANT) {
      ^
/Users/jsquyres/git/hwloc/include/hwloc/plugins.h:435:28: note: ‘filter’ was declared here
   enum hwloc_type_filter_e filter;
                            ^
```

Which is interesting, because the issue is in a .h file, but it only shows up when compiling one particular .c file (because the function in question that _might_ initialize the variable in question is in that .c file).  

In short:

``` c
hwloc_topology_get_type_filter(topology, type, &filter);
```

may actually leave `filter` uninitialized if `type` is out of range.  This commit simply ensures to initialize `filter` to `NONE` before calling `hwloc_topology_get_type_filter()`.

Signed-off-by: Jeff Squyres jsquyres@cisco.com

### DIFF
--- a/include/hwloc/plugins.h
+++ b/include/hwloc/plugins.h
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2013-2016 Inria.  All rights reserved.
+ * Copyright © 2016 Cisco Systems, Inc.  All rights reserved.
  * See COPYING in top-level directory.
  */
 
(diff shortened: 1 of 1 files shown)


bgoglin:
Looks like I pushed this to master 21 months ago as bd3bc912d7e4e3724c32bb8d54a2951190852ea1

PR #193, opened 2016-06-26
docs: print helpful error when "make dist" should fail

jsquyres:
If you try to run "make dist" when you don't have all the right
documentation generation tools, you currently get a misleading error
about a file being missing.  This commit adds a helpful message
specifically telling you that you are missing some tools that hwloc
needs, and therefore "make dist" will fail.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -1,6 +1,6 @@
 # Copyright © 2009-2016 Inria.  All rights reserved.
 # Copyright © 2009-2013 Université Bordeaux
-# Copyright © 2009-2014 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


bgoglin:
@jsquyres Can you look/try commit https://github.com/bgoglin/hwloc/commit/d6b8f4e3a3046d47443620a2ccef302510335c0d?
I had failures in jenkins when building with VPATH from a tarball with prebuilt doc on systems without doxygen. Adding !HWLOC_INSTALL_DOCS means that the error only appears when doxygen isn't available and there aren't prebuilt docs.

jsquyres:
Am checking...

Do you want a github webhook call to your Jenkins (so that it can build pull requests, etc.)?

bgoglin:
Yes, I was going to look at this but our Jenkins platform is under maintenance today.

PR #196, opened 2016-07-12
bind.c: enusre p is always initialized

jsquyres:
Fix compiler warning that p may be uninitialized.


jsquyres:
Disregard this PR, my v1.11 branch was accidentally way behind...

PR #200, opened 2016-07-21
Do not allocate zero size buffer.

cmcantalupo:
Signed-off-by: Christopher M. Cantalupo <christopher.m.cantalupo@intel.com>

### DIFF
--- a/hwloc/topology.c
+++ b/hwloc/topology.c
@@ -2215,15 +2215,17 @@ hwloc_build_level_from_list(struct hwloc_obj *first, struct hwloc_obj ***levelp)
   }
   nb = i;
 
-  /* allocate and fill level */
-  *levelp = malloc(nb * sizeof(struct hwloc_obj *));
-  obj = first;
-  i = 0;
-  while (obj) {
-    obj->logical_index = i;
-    (*levelp)[i] = obj;
-    i++;
-    obj = obj->next_cousin;
+  if (nb) {
+    /* allocate and fill level */
+    *levelp = malloc(nb * sizeof(struct hwloc_obj *));
+    obj = first;
+    i = 0;
+    while (obj) {
+      obj->logical_index = i;
+      (*levelp)[i] = obj;
+      i++;
+      obj = obj->next_cousin;
+    }
   }
 
   return nb;


cmcantalupo:
@bgoglin

I'm using hwloc in our project here:

https://github.com/geopm/geopm

We are getting a segfault in our tests on fedora 24:

https://build.opensuse.org/build/home:cmcantalupo/Fedora_24/x86_64/geopm/_log

Electric fence turned up the issue fixed by this patch and I think this is causing the segmentation fault in our tests.  If the patch looks good, please accept the pull request.

The error can be reproduced with the following command if hwloc is installed with the prefix $HOME/build/hwloc and the command is run in the git root.

LD_PRELOAD=/usr/lib64/libefence.so LD_LIBRARY_PATH=$HOME/build/hwloc/lib:$LD_LIBRARY_PATH tests/hwloc/.libs/hwloc_topology_dup

Thanks,
Chris

bgoglin:
Why does malloc(0) fail on Fedora 24? My glibc manual says that malloc(0) is valid.
Anyway, the patch is fine, but I don't see why it would cause a segfault. Your link to build.opensuse.org is unaccessible to anonymous users unfortunately.

cmcantalupo:
Agreed that malloc(0) is not strictly invalid, however, is it possible that the object created in this way is misused later in the code?  I'm just validating that the patch fixes the problem I am seeing on the geopm fedora 24 tests in OBS.  I'll get back here when I have the result.

cmcantalupo:
The patch does not change the geopm test failure on fedora 24, but it does stop electric fence from complaining.  Thanks for the feedback.

bgoglin:
Can you send more details about how to reproduce the failure or make the build log public ?

cmcantalupo:
Until I can prove otherwise I will assume the failure I'm looking into is due to a bug in geopm.  The hwloc tests only fail when electric fence is monitoring for malloc(0) occurrences.  The patch provided by this pull request does alleviate this issue, but as you mentioned, malloc(0) is not necessarily incorrect.

bgoglin:
I applied your patch to master and v1.11 branches, so I am closing this PR. If you find a deeper issue causing the GEOPM segfault, feel free to reopen and/or open a new PR or issue.
As a thank you, I added GEOPM to the list of hwloc users on the website :)

cmcantalupo:
Thanks!  Found the bug, in the end it was a bad interaction between googletest 1.7 and gcc6:

https://github.com/google/googletest/issues/705

The fix is here:

https://github.com/cmcantalupo/geopm/commit/bc1ba0bf883d3224fcaf811fd45bbb0ae23e1e09

PR #210, opened 2016-09-20
[RFC] New distances API for hwloc 2.0

bgoglin:
Everything is in the new dedicated header:
    https://github.com/open-mpi/hwloc/blob/distances/include/hwloc/distances.h

The main reason for this big change is that the old API was a pain in the internal implementation. Things are much more simple now.

Main changes:
* Distances are not public structures attached to the topology anymore. They have to be explicitly queried with hwloc_distances_get() functions.
* The public API manipulates objects (instead of indexes that can change, be duplicated, or be missing in corner cases).
* Distance values are uint64_t instead of floats. Users didn't like floats. Values are integer in the vast majority of cases (NUMA distances reported by the BIOS). uint64_t should give a wide-enough range in case users really need to input custom values. We could also support both integers and floats but I am not sure it's really useful.

The main drawback is that hwloc 1.x won't be able to import distances from XMLs generated by hwloc 2.x. This may be a problem if two components such as the resource manager and the MPI implementation use different hwloc releases. One idea is to add a flag to hwloc_topology_export_xml() to specify whether we want to export for hwloc 1.x or 2.x. Still to be discussed.


bgoglin:
@ggouaillardet IIRC you wondered why we used floats for distances. FWIw, we'll use uint64_t instead of float in hwloc 2.0.

PR #217, opened 2016-11-04
Just a couple of .gitignore fixes. Thanks!

cota:


bgoglin:
I merged those into a single commit with additional updates and pushed to master (with the wrong PR id), thanks!

PR #247, opened 2017-08-01
Fix evident lack of memory allocations handling in topology-linux.c

kutelev:
Due to not checking return values of some functions hwloc crashes in some cases when memory allocation (malloc) fails. To prevent such crashes there few checks have been added.

### DIFF
--- a/hwloc/topology-linux.c
+++ b/hwloc/topology-linux.c
@@ -803,6 +803,8 @@ hwloc_linux_set_tid_cpubind(hwloc_topology_t topology __hwloc_attribute_unused, 
 
   setsize = CPU_ALLOC_SIZE(last+1);
   plinux_set = CPU_ALLOC(last+1);
+  if (!plinux_set)
+    return -1;
 
   CPU_ZERO_S(setsize, plinux_set);
   hwloc_bitmap_foreach_begin(cpu, hwloc_set)
@@ -869,6 +871,10 @@ hwloc_linux_find_kernel_nr_cpus(hwloc_topology_t topology)
   fd = open("/sys/devices/system/cpu/possible", O_RDONLY); /* binding only supported in real fsroot, no need for data->root_fd */
   if (fd >= 0) {
     hwloc_bitmap_t possible_bitmap = hwloc_bitmap_alloc_full();
+    if (!possible_bitmap) {
+      close(fd);
+      return -1;
+    }
     if (hwloc__read_fd_as_cpulist(fd, possible_bitmap) == 0) {
       int max_possible = hwloc_bitmap_last(possible_bitmap);
       hwloc_debug_bitmap("possible CPUs are %s\n", possible_bitmap);
@@ -882,6 +888,8 @@ hwloc_linux_find_kernel_nr_cpus(hwloc_topology_t topology)
 
   while (1) {
     cpu_set_t *set = CPU_ALLOC(nr_cpus);
+    if (!set)
+      return -1;
     size_t setsize = CPU_ALLOC_SIZE(nr_cpus);
     int err = sched_getaffinity(0, setsize, set); /* always works, unless setsize is too small */
     CPU_FREE(set);
@@ -911,8 +919,12 @@ hwloc_linux_get_tid_cpubind(hwloc_topology_t topology __hwloc_attribute_unused, 
 
   /* find the kernel nr_cpus so as to use a large enough cpu_set size */
   kernel_nr_cpus = hwloc_linux_find_kernel_nr_cpus(topology);
+  if (kernel_nr_cpus == -1)
+    return -1;
   setsize = CPU_ALLOC_SIZE(kernel_nr_cpus);
   plinux_set = CPU_ALLOC(kernel_nr_cpus);
+  if (!plinux_set)
+    return -1;
 
   err = sched_getaffinity(tid, setsize, plinux_set);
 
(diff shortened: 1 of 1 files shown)


hwloc/topology-linux.c
```diff
@@ -869,6 +871,8 @@ hwloc_linux_find_kernel_nr_cpus(hwloc_topology_t topology)
   fd = open("/sys/devices/system/cpu/possible", O_RDONLY); /* binding only supported in real fsroot, no need for data->root_fd */
   if (fd >= 0) {
     hwloc_bitmap_t possible_bitmap = hwloc_bitmap_alloc_full();
+    if (!possible_bitmap)
+      return -1;
```
sthibaul: fd needs to be closed before returning

hwloc/topology-linux.c
```diff
@@ -1851,9 +1875,11 @@ hwloc_linux_get_area_membind(hwloc_topology_t topology, const void *addr, size_t
   int pagesize = hwloc_getpagesize();
   char *tmpaddr;
   int err;
-  unsigned i;
+  int i;
```
sthibaul: This seems useless?

bgoglin:
Just for fun: Did you actually observe those exact crashes? There are other places where we don't check allocations but I am too lazy to fix all of them...

kutelev:
Yes, we really observed these crashes and to be honest only places where it was really crashing have been fixed.

hwloc/topology-linux.c
kutelev: Reverted.

hwloc/topology-linux.c
kutelev: Thanks for pointing an issue. Now should be fixed.

bgoglin:
I'd like to reproduce that to make sure we don't need other more important allocation checks later in the discovery process in topology.c. Did you use a fuzzer to generate false allocation failures? Or did you really fill your entire address space to make malloc fail?

bgoglin:
Please squash your commits into a single one (and push -f) so that the review process doesn't appear in the git history. And please revert the other int/unsigned changes, those are unrelated to malloc failures.

kutelev:
hwloc_linux_find_kernel_max_numnodes now returns -1 on error that is why max_os_index type was changed from unsigned to int. Any suggestion how hwloc_linux_find_kernel_max_numnodes can report an error without returning -1? Thanks in advance.

bgoglin:
Just testing whether max_os_index == (unsigned)-1 should work fine without changing the non-error case (that piece of code is hard to test for regressions).

kutelev:
Regarding to the reproducibility in tests allocations are failed synthetically using simple library which is preloaded using LD_PRELOAD and DYLD_INSERT_LIBRARIES. Allocations are failed randomly using different strategies to cover as much as possible cases inclusing completelly abnormal. But in any case due to strict requirements in our case all situations must be handled. Application shall never crash.

bgoglin:
OK that's what I called a fuzzer. I am surprised that no other problem appeared outside of this file but I am not going to complain about it.

kutelev:
Such crazy test exists only for Linux and Mac OS X now. Windows in plans. Maybe other issues will be found there later.

kutelev:
Single commit has been kept. All int/unsigned changes have been reverted.

bgoglin:
Thanks, I'll merge this to master and v1.11 once the testing infrastructure is fixed (hopefully before the end of the week).
If you can share your way to perform this malloc fuzzing testing and maybe even some way to automatize it, that'd be nice.

kutelev:
Thank You. About sharing the approach I will be honest and will not promise that I will provide the working solution immediately. In fact all I can share is library which fails allocations when preloaded. Still not sure that there is high value in this library, it contains ~200 lines of quite simple code. If you still interested in it I can prepare an example application which is tested using such approach, but it will take some time in any case.

mofr:
Why using `(unsigned)-1` was considered as a better approach?

kutelev:
After digging deeper we found (as was predicted) other places where hwloc fails. For example hwloc_topology_init crashes in some cases. As there are more failures that we may want to see simple project which can demonstrate the problem has been created. You can find it there: https://github.com/kutelev/hwlocoverthrower/tree/master/hwlocoverthrower. In the .travis.yml file you can sequence of command using which problem can be reproduced. It uses a simple library (overthrower) which was mentioned earlier. Test application itself is extremely simple, all it does is:
```c
for (i = 0; i < 1000; ++i) {
    if (hwloc_topology_init(&topology) != 0)
        continue;
    hwloc_topology_destroy(topology);
}
```
When overthrower is not preloaded this code finishes correctly, but when is overthrower preloaded this code crashes with probability which is close to 100%. Build on Travis CI can be found there: https://travis-ci.org/kutelev/hwlocoverthrower. Now I shared all I can.

bgoglin:
hwloc_topology_init() would be easy to fix. If you had load() between init() and destroy(), it would be a lot harder.

I can't find the code for liboverthrower. But we could use something else like libfuzzer anyway.

kutelev:
In real life we have hwloc_topology_load after hwloc_topology_init. Regarding to the liboverthrower it is stored in separate a repository.

Full list of commands which can be executed to retrieve and all necessary dependencies:
```bash
git clone https://github.com/kutelev/hwlocoverthrower.git
cd hwlocoverthrower
git clone --depth 1 https://github.com/open-mpi/hwloc.git
git clone --depth 1 https://github.com/kutelev/overthrower.git
cd hwloc
./autogen.sh
./configure --disable-libxml2 --enable-static --disable-shared --disable-libudev --disable-pci
make
cd ../overthrower
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=1 .
cmake --build . --config Release
cd ../hwlocoverthrower
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=1 .
cmake --build . --config Release
cd ..
python ./tests/test.py
```

bgoglin:
Replaced by #248

PR #273, opened 2017-11-07
update lstopo graphical output to memory children

bgoglin:
NUMA nodes don't contain packages/cores/caches/PUs anymore. They don't have a double box anymore. They are on the side, just like other PUs.

To simplify reading in most cases, we actually display memory children above other children (unless --children-order=plain is given). When the parent is a cache, we even display memory children above that cache.

Example with a fake mini-KNL in SNC4/Hybrid:
knl-hybrid


bgoglin:
Pushed as ea4537a51cc565fbeda71c30571fd460f34e88d1

PR #286, opened 2018-01-23
Consistent naming of Intel(R) Xeon Phi(TM) processors in messages and documentation

dawid-lukwinski:
For both Knights Landing (KNL) and Knights Mill (KNM) processors mostly
the same code and documentation is used. Therefore text regarding these
processors has to be consistent and appropriate for both KNL and KNM.

Signed-off-by: Dawid Łukwiński <dawid.lukwinski@intel.com>

### DIFF
--- a/utils/hwloc/hwloc-dump-hwdata-knl.c
+++ b/utils/hwloc/hwloc-dump-hwdata-knl.c
@@ -18,7 +18,7 @@
 
 #define KERNEL_SMBIOS_SYSFS "/sys/firmware/dmi/entries"
 
-#define KNL_SMBIOS_GROUP_STRING "Group: Knights Landing Information"
+#define KNL_SMBIOS_GROUP_STRING "Group: Intel(R) Xeon Phi(TM) Processor Information"
 
 /* Header is common part of all SMBIOS entries */
 struct smbios_header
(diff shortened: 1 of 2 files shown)


utils/hwloc/hwloc-dump-hwdata-knl.c
```diff
@@ -18,7 +18,7 @@
 
 #define KERNEL_SMBIOS_SYSFS "/sys/firmware/dmi/entries"
 
-#define KNL_SMBIOS_GROUP_STRING "Group: Knights Landing Information"
+#define KNL_SMBIOS_GROUP_STRING "Group: Intel(R) Xeon Phi(TM) Processor Information"
```
bgoglin: This string is supposed to be read from SMBIOS tables, I don't think we can change it without breaking things.
But I can't make that change break our regression tests, hwloc-dump-hwdata still finds the KNL SMBIOS group.
@grzegorz-andrejczuk Can you confirm?

bgoglin:
By the way, I pushed a big doc update this morning with at least another instance of "KNL" in doc/hwloc.doxy

I have to say I am not a big fan of writing "Intel(R) Xeon Phi(TM)" everywhere, even in debug messages from tools that automatically run during boot without any human ever looking at the output.

utils/hwloc/hwloc-dump-hwdata-knl.c
grzegorz-andrejczuk: Yes, these strings are part of DMI entries definition and should not be changed. When the string do not match MCDRAM cache will not be detected. 
Good catch, Thanks.

bgoglin: There might be an issue in the current code. When I change the #define'd string, strncmp() doesn't match it anymore, but is_knl_entry() still returns 1: it skips the KNL string that I modified on purpose, and then it tries again on the remaining string, which is now empty, hence strncmp() now matches because we pass len=0.

PR #288, opened 2018-01-29
hwloc-dump-hwdata: fix the lookup for KNL SMBIOS entries

grzegorz-andrejczuk:
We always matched because the last iteration would strncmp() with
len=0.

Signed-off-by: Grzegorz Andrejczuk <grzegorz.andrejczuk@intel.com>


bgoglin:
Applied, thanks!

PR #289, opened 2018-01-31
KNM (Knights Mill) SMBIOS detection fix.

dawid-lukwinski:
KNM can use the same code as KNL, except the different SMBIOS string.
This commit adds a separate SMBIOS group string for KNM.


bgoglin:
Applied, thanks!

PR #295, opened 2018-02-09
hwloc-dump-hwdata: add Dell SMBIOS Group string

grzegorz-andrejczuk:
This commit makes SMBIOS groups string easier to extend.
This way it will be possible to easily add requested vendor
string to be matched in SMBIOS Group.

SMBIOS was replaced with vendor name in existing strings,
because whole name might be too long.

Signed-off-by: Grzegorz Andrejczuk <grzegorz.andrejczuk@intel.com>

### DIFF
--- a/utils/hwloc/hwloc-dump-hwdata-knl.c
+++ b/utils/hwloc/hwloc-dump-hwdata-knl.c
@@ -18,8 +18,16 @@
 
 #define KERNEL_SMBIOS_SYSFS "/sys/firmware/dmi/entries"
 
-#define KNL_SMBIOS_GROUP_STRING "Group: Knights Landing Information"
-#define KNM_SMBIOS_GROUP_STRING "Group: Knights Mill Information"
+#define KNL_INTEL_GROUP_STRING "Group: Knights Landing Information"
+#define KNM_INTEL_GROUP_STRING "Group: Knights Mill Information"
+#define KNL_DELL_GROUP_STRING "Knights Landing Association"
+
+static char *allowed_group_strings[] =
+{
+    KNL_INTEL_GROUP_STRING,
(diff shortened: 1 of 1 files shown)


bgoglin:
It works on both Dell BIOS, thanks. I'll apply it and send a mail to users for testing on other vendors' platforms. Do you only have Intel-specific chassis/motherboards/BIOS, or do you also test on vendors' product?

grzegorz-andrejczuk:
I do not have machines from other vendors. I have tested this on Intel only.

bgoglin:
How about we add some printf to ease debugging (and addition of other vendor-specific strings). We'd see this when matched:
```
Dumping KNL SMBIOS Memory-Side Cache information:
  File = /home/bgoglin/Bureau/hwloc/knl-snc4h50//sys/firmware/dmi/entries/14-0/raw, size = 4096
    Looking for "Group: Knights Landing Information" in group string "Group: Knights Landing Information"
   Found phi group
  Found KNL type = 160
[...]
```
and this when not matched (I changed the strings to emulate a failure):
```
Dumping KNL SMBIOS Memory-Side Cache information:
  File = /home/bgoglin/Bureau/hwloc/knl4//sys/firmware/dmi/entries/14-0/raw, size = 4096
    Looking for "Group: Knights Landing Information" in group string "Knights Landing Association"
    Looking for "Group: Knights Mill Information" in group string "Knights Landing Association"
    Looking for "Knights Landing Associatiooooon" in group string "Knights Landing Association"
  Failed to find phi group
SMBIOS table does not contain KNL entries
```

grzegorz-andrejczuk:
I added logging, fixed off-by-one error and created new PR
https://github.com/open-mpi/hwloc/pull/296

PR #311, opened 2018-05-08
hwloc/rename: re-add components symbols

evan-charmworks:
These definitions were originally removed in git commit 93abf09fee121c55b99f578d62e3ea21decdfbed, aka SVN r5097.

They are needed to avoid symbol conflicts, for example between Open MPI's copy of hwloc and something that uses both Open MPI and its own copy of hwloc. Conflicts can occur both at binary link time and at runtime, during dynamic component lookup.

If this is acceptable, I can create PRs for the v1.11 and v2.0 branches as well.


jsquyres:
Per https://github.com/open-mpi/ompi/wiki/Admistrative-rules#contributors-declaration, can you add a signed-off-by line in your commit message (use `git commit -s`)?  This line indicates your agreement with our contributor's declaration.

evan-charmworks:
Signed-off-by added, sorry about that.

evan-charmworks:
Rationale added to commit message.

bgoglin:
What bothers me is that your patch will likely break make check under tests/hwloc/rename (because those symbols aren't defined in any header, and that test looks for missing and unneeded renames).

I need to think about a proper way to make this work without having to hardwire the whitelist in the test.

Do you actually see a real-life issue without your patch? I am trying to see why we never had this issue before.

evan-charmworks:
Symbol conflicts were reported by a user of Charm++, which uses hwloc. You can see the thread here: https://lists.cs.illinois.edu/lists/arc/charm/2018-05/msg00000.html

bgoglin:
I combined your patch with some moving of declarations and pushed it to my fork for testing.
https://github.com/bgoglin/hwloc/commit/d0c2a33dfdff954e7042804db53c027f23ea7069
The only non-trivial difference with your rename.h is the removal of hwloc_fake_components (it's never linked with libhwloc, it's only built for tests).

I think your patch is enough for Charm++ since you're embedding. My additional changes shouldn't matter to you.

If that's OK for you, I'll push my patch to master and stable branches later tonight.

Thanks for your work. By the way, good to know that Charm++ is also using embedding.

evan-charmworks:
Sounds good to me. Thanks for looking at this.

evan-charmworks:
If it's any help, this is the one-liner I used to generate the definitions for my commit:

`grep -Eroh "hwloc_component hwloc_[a-zA-Z0-9_]+_component" | sort | uniq | awk '{print "#define " $NF " HWLOC_NAME(" $NF ")"}' | sed 's/(hwloc_/(/'`

bgoglin:
@evan-charmworks Do you need an official release with this fix in the near future? It's the only user-visible fix since 1.11.10 and 2.0.0 were released, I'd like to know if it's enough to deserve new stable releases. If you're OK with the patch on top of your embedded hwloc, I'll wait for more fixes before releasing.

evan-charmworks:
Charm++ will be fine with the patch to our embedded copy of hwloc. No need to push a release for our sake. Thanks!

PR #317, opened 2018-07-16
Check for freed memory before using it

wesbland:
The error case of `hwloc_convert_from_v1dist_floats` cleans up memory
which might have previously been freed if the right set of error
branches occur. By explicitly setting the memory pointer to `NULL` after
freeing it, we can later check that value and make sure we don't
dereference an freed pointer.

See Coverity defects 185552, 185546, and 185542 in MPICH-CH4 project

### DIFF
--- a/hwloc/topology-xml.c
+++ b/hwloc/topology-xml.c
@@ -1777,20 +1777,22 @@ hwloc_convert_from_v1dist_floats(topology, nbobjs, v1dist->floats, values);
     fprintf(stderr, "%s: XML component discovery failed.\n",
 	    data->msgprefix);
  err:
-  hwloc_free_object_siblings_and_children(root->first_child);
-  root->first_child = NULL;
-  hwloc_free_object_siblings_and_children(root->memory_first_child);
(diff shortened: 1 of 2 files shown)


wesbland:
I'm sure there's a commit prefix that I should be using here, but I couldn't figure out the scheme. I'm happy to update the patch if you can give me a pointer in the right direction.

bgoglin:
I requested access to ch4's coverity since hwloc's coverity doesn't complain (and I don't understand the issue so far).

bgoglin:
185546 and 185552 are false positives, root and root->nodeset are only freed in hwloc__xml_import_object() when the input object isn't the root (parent!=NULL).

185542 is also a false positive. The object is freed in hwloc__insert_object_by_cpuset() only when insertion fails (return NULL) or the object gets merged with another one (return != obj). Both these conditions are verified before dereferencing the object again.

They were already marked as false-positive in hwloc's coverity. I guess you can ignore everything hwloc related in MPICH's coverity.

wesbland:
Sounds good. I'll clear those out then.

PR #324, opened 2018-08-14
Define HWLOC_VERSION* in hwloc/autogen/config.h public header

bgoglin:
OMPI would like a way to check the hwloc library version (not API version) from external configure scripts (open-mpi/ompi#5395). We currently have HWLOC_VERSION defined in private/autogen/config.h
It's only there because it wasn't prefixed with HWLOC_ in the past. Now that it's properly prefixed, there's no reason to keep it private.
Also add HWLOC_VERSION_{MAJOR,MINOR,RELEASE,GREEK} to avoid external parsing.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -5,7 +5,7 @@
 # major, minor, and release are generally combined in the form
 # <major>.<minor>.<release>.  If release is zero, then it is omitted.
 
-# Please update HWLOC_VERSION in contrib/windows/private_config.h too.
+# Please update HWLOC_VERSION* in contrib/windows/hwloc_config.h too.
 
 major=2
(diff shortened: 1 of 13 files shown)


bgoglin:
@ggouaillardet Do you want this in hwloc 2.0.2? My understanding is that there's no need to put this in hwloc 1.11.11 since OMPI will never use the system-wide 1.11.x unless forced to.

ggouaillardet:
@bgoglin thanks a lot !
as you said, starting from hwloc 2.0.2 is far enough
(and I will add a `#ifdef HWLOC_VERSION` this is only tested when available)

PR #342, opened 2018-12-14
Fix documentation on API version

hannesweisbach:
The documentation used a right-shift operator to construct HWLOC_API_VERSION from a release version. This made no sense and the left-shift operator should be used.

Signed-off-by: Hannes Weisbach hannes.weisbach@gmail.com


bgoglin:
Thanks, I found another similar typo in the doxy file, I'll fix it.

PR #345, opened 2018-12-19
linux: remove /proc/cpuinfo-based topology detection

bgoglin:
sysfs topology is available since 2.6.16,
stop parsing core/package information in /proc/cpuinfo

/proc/cpuinfo is still used for getting CPUModel etc.

Signed-off-by: Brice Goglin <Brice.Goglin@inria.fr>


bgoglin:
I'll likely merge this in 2.1 unless we find an architecture where /proc/cpuinfo still has more info than sysfs.

PR #357, opened 2019-03-25
[WIP DNM] Ignore NVIDIA GPU memory NUMA nodes by default

bgoglin:
Remove them from the topology so that we don't allocate there by mistake. Another option would be to mark them as disallowed by default.

If we keep them, we should say which GPU corresponds to which NUMA node (subtype + PCIBusID info attr).

TODO add a test

We're not supposed to have NUMA nodes inside I/O devices right now. We fixed the locality to attach to the correct socket, but not to the GPU itself.

We may have a similar question when we'll have NVDIMM-like persistent memory regions on future NVMe.


bgoglin:
Merged in master 4c37b1d086704bc6bc75b052650fb196fa896791

PR #359, opened 2019-05-22
hwloc-affinity

NicolasDenoyelle:
Extend hwloc utilities with a tool to bind threads to single processing units with a policy.


sthibaul:
Interesting :)

In cpuaffinity_exec, I believe you need to add a synchronization between the child and the parent. Otherwise the child might have the time to clone() before the parent manages to call ptrace().

bgoglin:
Options -i and --input should use the existing helpers in utils.h so that they handle XML/CPUID/Synthetic etc.

There's also some duplication in hwloc_obj_from_string() which already exists as generic helpers in hwloc-calc.h with support for ranges etc.

If it's only needed for --restrict obj:X, we should extend --restrict in other tools to also support object specifications. Things should be uniform there. Unfortunately, once you start doing that, it means you need to support both physical and logical specification with command-line options. I wonder if we should just use --restrict cpuset and keep telling people to use hwloc-calc for generating the cpuset. Or we should support foo:Px and foo:Lx for logical and physical indexes.

The scatter policy already exists in hwloc-distrib, which could also be extended to export objects and/or indexes.

I think this should be split into two parts:
1) one for generating list of objects, and this should be integrated in hwloc-calc. It's just a matter of adding new output options. We already have options for exporting with different separators, physical or logical index, etc. Those will be needed for your code anyway.
Also -i and --input are already supported there.
2) one for binding new threads using such a list, which doesn't have -i or --input since it uses binding

NicolasDenoyelle:
> Options -i and --input should use the existing helpers in utils.h so that they handle XML/CPUID/Synthetic etc.

Done 

> There's also some duplication in hwloc_obj_from_string() which already exists as generic helpers in hwloc-calc.h with support for ranges etc.

Painfully done :)

> If it's only needed for --restrict obj:X, we should extend --restrict in other tools to also support object specifications. Things should be uniform there. Unfortunately, once you start doing that, it means you need to support both physical and logical specification with command-line options. I wonder if we should just use --restrict cpuset and keep telling people to use hwloc-calc for generating the cpuset. Or we should support foo:Px and foo:Lx for logical and physical indexes.

Done, logical option will work both for input and output. 
As a user myself I enjoy the feature allowing to name object and indexes. It significantly shorten command lines. I referred to hwloc-calc(1) for details about input format.

> The scatter policy already exists in hwloc-distrib, which could also be extended to export objects and/or indexes.
> 
> I think this should be split into two parts:

Yet to be done

NicolasDenoyelle:
> Interesting :)
> 
> In cpuaffinity_exec, I believe you need to add a synchronization between the child and the parent. Otherwise the child might have the time to clone() before the parent manages to call ptrace().

I tried to fix this by attaching inside child instead of parent before calling execvp. However, I remember having issue with way of doing it. No errors are showing up but it should be tested.

NicolasDenoyelle:
> The scatter policy already exists in hwloc-distrib, which could also be extended to export objects and/or indexes.

I looked at the code at of hwloc-distrib. It is mainly based on hwloc_distrib() function in helper.h
I think a good think to do would be to update this function to allow to implement the tleaf as it is done
in hwloc-tbind-obj.c. It does a very similar work. The interface can nearly remains the same (instead of providing a starting and ending depths, provide a list of depths) and allow for distribution policies.

NicolasDenoyelle:
> Interesting :)
> 
> In cpuaffinity_exec, I believe you need to add a synchronization between the child and the parent. Otherwise the child might have the time to clone() before the parent manages to call ptrace().

I think this is properly fixed now with synchronization through signals.

sthibaul:
> > Interesting :)
> > In cpuaffinity_exec, I believe you need to add a synchronization between the child and the parent. Otherwise the child might have the time to clone() before the parent manages to call ptrace().
> 
> I think this is properly fixed now with synchronization through signals.

Perhaps using semaphores would be simpler?

jsquyres:
Did something go wrong with this PR?  There are currently 256 commits listed here...?

bgoglin:
@jsquyres I'd say many fixup commits are not squashed, and most of them are duplicated because the branch was merged with a rebased version of itself.
@NicolasDenoyelle Note that PRs are automatically updated if rebase/squash and push --force to cleanup your history

bgoglin:
You should likely split this PR in small and easier-to-review pieces to ease its converging/merging. One first good step would be to just add hwloc-bind-threads, with this as a possible usage:
```
hwloc-bind-threads -t core -- myprog
```
Where -t core means we're binding the first thread to first core, second thread to second core, etc. And last threads are not bound if there aren't enough cores. Or option to wrap-around.
```
hwloc-bind-threads <location> ... -- myprog
```
This would bind each thread to its respective location if any, where location is implemented exactly as in hwloc-calc/bind/...

Add --single and --strict just like in hwloc-bind.

Unless I am mistaken, Linux is the only OS where a process can bind a thread of another process, hence all this can be made Linux-specific.

And add a basic test that counts the number of cores, spawns N+1 threads with pthread_create and they check their own binding once ptrace had time to bind them. Not sure testing with OpenMP is a good idea because the way OS threads are created and given to user OpenMP tasks can vary from one implementation to another.

sthibaul:
Actually I believe windows can as well, by using OpenThread to get a handle on the thread to be bound

config/ax_pthread.m4
```diff
@@ -0,0 +1,486 @@
...
+#   scripts that are the output of Autoconf when processing the Macro. You
+#   need not follow the terms of the GNU General Public License when using
+#   or distributing such scripts, even though portions of the text of the
+#   Macro appear in them. The GNU General Public License (GPL) does govern
+#   all other use of the material that constitutes the Autoconf Macro.
+#
+#   This special exception to the GPL applies to versions of the Autoconf
+#   Macro released by the Autoconf Archive. When you make and distribute a
+#   modified version of the Autoconf Macro, you may extend this special
+#   exception to the GPL to apply to your modified version as well.
```
jsquyres: This file might be a deal-breaker for hwloc.

hwloc is distributed under the BSD license.  This file specifically states that it is GPL v3 or later.  GPL v3 is... difficult to work with.

There are the "special exceptions", but that seems to only cover "*the configure scripts that are the output of Autoconf when processing the Macro*."  To me (and I am *not* a lawyer / this is not legal advice / yadda yadda yadda), this sounds like we would then have a GPL v3 file in our repo (ouch).  I don't recall offhand if hwloc includes its `.m4` files in distribution tarballs (I think it does?), which would mean that there is a GPL v3 file in an otherwise-BSD tarball.  That gets tricky to explain to people.

I don't know if the above analysis is correct or not, but why even introduce the question?  Why raise the spectre of GPL v3?

My $0.02: just write the tests yourself and stay away from the AX stuff.

config/hwloc_internal.m4
```diff
@@ -355,6 +355,32 @@ EOF
...
+       #include "sys/syscall.h"
+       #ifndef SYS_gettid
+       #error "syscall SYS_gettid not found"
+       #endif
+       int main(void){ return syscall(SYS_gettid) > 0;}
+    ]])], AC_DEFINE([HWLOC_HAVE_SYS_GETTID], [1], [syscall header is present and SYS_gettid macro is defined]))
+    
+
+
+
```
jsquyres: Lots of excess blank lines here; please remove.

config/hwloc_internal.m4
```diff
@@ -355,6 +355,32 @@ EOF
...
+    AC_CHECK_HEADERS([sys/ptrace.h], [
+      AC_CHECK_FUNCS([ptrace],[hwloc_have_ptrace=yes], [hwloc_have_ptrace=no])
+      AC_COMPILE_IFELSE([AC_LANG_SOURCE([[ 
+        #include "sys/ptrace.h"
+        int main(void){ return ptrace(PTRACE_SEIZE, -1, 0, (void*)(PTRACE_O_TRACECLONE|PTRACE_O_TRACEFORK)); }
+      ]])],, [hwloc_have_ptrace=no])
+    ])
+    AM_CONDITIONAL([HWLOC_HAVE_PTRACE],[ test "x$hwloc_have_ptrace" = xyes ])
+    if test x$hwloc_have_ptrace = xyes; then
+       AC_DEFINE([HWLOC_HAVE_PTRACE], [1], [ptrace is present and supportes PTRACE_SEIZE])
```
jsquyres: It is safer to *always* `AC_DEFINE[_UNQUOTED]` a value -- i.e., always define it to 0 or 1 (vs. undef or 1).  

Specifically, if you go the "undef or 1" route, your C code looks like this:

```c
#ifdef HWLOC_HAVF_PTRACE
// do things with ptrace
```

But if you go the "always DEFINE to 0 or 1" route, your C code looks like this:

```c
#if HWLOC_HAVF_PTRACE
// do things with ptrace
```

I.e., `#if` vs. `#ifdef`.

The reason `#if` is better is because the compiler/preprocessor will tell you if you have a typo in your name, whereas `#ifdef` will not.

Did you notice that I misspelled `HWLOC_HAVE_PTRACE`, above?  The first code -- with `#ifdef` -- will silently always be false.  The second code -- with `#if` -- will give a compiler warning about an undefined preprocessor macro (`HWLOC_HAVF_PTRACE`), thereby alerting you to your typo.  Defensive coding === good.

Meaning: please always define logical preprocessor macros to 0 or 1 -- not undef-vs-1.

Thanks.

config/hwloc_internal.m4
```diff
@@ -385,7 +411,12 @@ AC_DEFUN([HWLOC_SETUP_TESTS],[
 ###
 EOF
 
-    AC_CHECK_LIB([pthread], [pthread_self], [hwloc_have_pthread=yes])
+    # Check thread support.
+    AX_PTHREAD([
+	hwloc_have_pthread=yes;
+	AC_DEFINE([HAVE_PTHREAD], [1], [pthread support is here])
```
jsquyres: Same as mentioned above:

* `AX_PTHREAD` could be problematic.  Are you just needing to see if you need to use `-pthread`?  If so, you might be able to steal Open MPI's pthread m4.
* Always define to 0 or 1.

utils/hwloc/Makefile.am
```diff
@@ -66,19 +67,37 @@ hwloc_ps_SOURCES = hwloc-ps.c
...
 
+noinst_LTLIBRARIES+=libhwloc-tbind.la
+libhwloc_tbind_la_SOURCES=hwloc-tbind-obj.c hwloc-tbind.h
+LIBDADD=$(HWLOC_top_builddir)/hwloc/libhwloc.la
+
+hwloc_tbind_SOURCES=hwloc-tbind.c hwloc-tbind.h
+LDADD+=$(HWLOC_top_builddir)/hwloc/libhwloc.la
+LDADD+=libhwloc-tbind.la
+
+AM_CFLAGS+=-DTBIND_XML_DIR=\"$(HWLOC_top_builddir)/tests/hwloc/xml\"
```
jsquyres: `-D` options should go in to `AM_CPPFLAGS`.

utils/hwloc/hwloc-tbind.h
```diff
@@ -0,0 +1,314 @@
+/***************************************************************************
+ * Copyright 2019 UChicago Argonne, LLC.
+ * Author: Nicolas Denoyelle
+ * SPDX-License-Identifier: BSD-3-Clause
```
jsquyres: Please put in the following line in your headers in new files (per other files in this repo):

```
 * See COPYING in top-level directory.
```

config/ax_pthread.m4
sthibaul: This question holds for all autoconf bits, it's not just ax_pthread.m4. See for instance config/libtool.m4, it's the same kind of text.

Put another way, we already do distribute gpl files in the hwloc tarballs since long :)

jsquyres: No, it's still different: libtool.m4 is not in our repo.

sthibaul: But it's distributed in our tarballs.
Actually, we could even not put ax_pthread.m4 in the repo, and assume that people running autogen.sh have autoconf-archive available.

jsquyres: I'm admittedly taking a fine line here between in the repo / generated at autogen.sh time...

But GPL3 is a whole spectre in itself.  Suffice it to say that GPL3 will... be difficult.  I don't know what the restrictions are, but ... I don't even want to find out.  😄 

I also don't think it's a good assumption that people will have autoconf-archive already available (e.g., I don't).  It would be another requirement.

There's also downstream considerations -- Open MPI, MPICH, etc.  Anyone who embeds hwloc.

Let me take a step back and ask: what exactly is the need for `AX_PTHREAD`?  Can't it be re-written in a simple test without raising any of these questions?

sthibaul: But I mean the question is already raised by the presence of libool.m4, which is definitely GPLv2 with the same exception as ax_pthread.m4

jsquyres: Let me take a step back and ask: what exactly is the need for `AX_PTHREAD`? Can't it be re-written in a simple test without raising any of these questions?

NicolasDenoyelle: AX_PTHREAD is for detection of pthread capabilities and defining macros to setup CFLAGS and LDFLAGS for compilation. In my opinion it is not a big deal to remove AX_PTHREAD and 
make a simple replacement instead. I assumed that all the boring portability checks would have been thoroughly performed there and I would not have reinvent the will. I didn't check the license though.. I am going to replace it with something more convenient licence-wise.

config/hwloc_internal.m4
NicolasDenoyelle: Thanks for the tip. I will change that.

utils/hwloc/hwloc-tbind.h
NicolasDenoyelle: Done see#https://github.com/open-mpi/hwloc/pull/359/commits/028c2eeaab0ba915743f068c85e5d4a8b0cfeaf5

utils/hwloc/Makefile.am
NicolasDenoyelle: Done.  see #https://github.com/open-mpi/hwloc/pull/359/commits/a40b3ebdf08136431805ac871ca6f895a3452d64

config/hwloc_internal.m4
NicolasDenoyelle: Done. See #https://github.com/open-mpi/hwloc/pull/359/commits/a7979a9c2aa3fd277a38c239432dcfb8db97202b

config/hwloc_internal.m4
NicolasDenoyelle: Done see #https://github.com/open-mpi/hwloc/pull/359/commits/a5a097d6cc8f762fbd2d32a88d9f977c61f822e9

config/hwloc_internal.m4
NicolasDenoyelle: fixed.

config/ax_pthread.m4
NicolasDenoyelle: I updated with the original check from hwloc master.
See #https://github.com/open-mpi/hwloc/pull/359/commits/a7979a9c2aa3fd277a38c239432dcfb8db97202b

NicolasDenoyelle: Please hit resolve if it suits you.

NicolasDenoyelle:
> Actually I believe windows can as well, by using OpenThread to get a handle on the thread to be bound
In order to call OpenThread, you have to know which thread to open. ptrace allow to hook on threads
creation. I can't find such a thing on windows documentation.
I found this: https://docs.microsoft.com/en-us/windows/desktop/Dlls/dllmain
which will notify you when a DLL will create a thread. However you can't get the thread ID, suspend it

bgoglin:
Forget about Windows for now, there are too many missing pieces there. Make all your thread-binding stuff Linux-specific as I said 17 days ago.

jsquyres:
Thanks for all the updates.

Taking a step back, though, I see a lot of other pthread tests in configure already, and I see pthread code used throughout the rest of hwloc.  Is there really a need for your new configure test?  Or does your new pthread-using C code just need to be protected by existing macros?

NicolasDenoyelle:
> it looks like this test is only run from hwloc's `configure.ac`. Is this macro also defined if hwloc is embedded?

ptrace configure tests are running in the scope of macros used for hwloc utilities. 
pthread tests are running indeed from `configure.ac`. I guess it does run in embedded mode.
From `configure.ac` and `config/hwloc_internal.m4` I could not find anything preventing it. 

> Is there really a need for your new configure test? Or does your new pthread-using C code just need to be protected by existing macros?

hwloc does not define macro HAVE_PTHREAD. pthread is required to enable some tests at make check. If it is not defined then tests do not run. I think there is a need for this test in order to 
have pthread-using C protected by this macro which has a different meaning than existing C macros.

>  Make all your thread-binding stuff Linux-specific as I said 17 days ago.

I am working on it. I'll do one branche and one PR per feature as you proposed.

jsquyres:
@NicolasDenoyelle Thanks for investigating the configury stuff (embedded, etc.).

bgoglin:
Note to open pull requests: some things changed in the CI yesterday, you'll need to rebase on top of master to avoid total CI failure.

Issue #373, opened 2019-12-09
UBSAN error: topology.c:2969:2: runtime error: null pointer passed as argument 2, which is declared to never be null


senior7515:
### What version of hwloc are you using?
2.1

* Run `which lstopo` to ensure that you are running the desired hwloc installation

I am integrating as a library. 

* Run `lstopo --version` to find their hwloc version

```
./lstopo --version
lstopo 2.1.0

```

  * If hwloc was built from a Git clone, report the Git commit hash from `git show`

```
md5sum hwloc-2.1.0.tar.gz 
f715051fb660549ef1278ad2545edd71  hwloc-2.1.0.tar.gz
```

### Which operating system and hardware are you running on?

* On Unix-like systems, run `uname -a` so that we know which operating system, distribution, and kernel version you are using.
```
Linux mountdoom 5.3.11-200.fc30.x86_64 #1 SMP Tue Nov 12 19:25:25 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

```

* Post the output of `lstopo -` if it works

```
/home/agallego/workspace/v/build/debug/v_deps_build/hwloc-prefix/src/hwloc/hwloc/topology.c:2969:2: runtime error: null pointer passed as argument 2, which is declared to never be null
/home/agallego/workspace/v/build/debug/v_deps_build/libpciaccess-prefix/src/libpciaccess/src/common_device_name.c:163:27: runtime error: shift exponent 4318 is too large for 32-bit type 'int'
Machine (63GB total)
  Package L#0
    NUMANode L#0 (P#0 63GB)
    L3 L#0 (8192KB)
      L2 L#0 (256KB) + L1d L#0 (32KB) + L1i L#0 (32KB) + Core L#0
        PU L#0 (P#0)
        PU L#1 (P#4)
      L2 L#1 (256KB) + L1d L#1 (32KB) + L1i L#1 (32KB) + Core L#1
        PU L#2 (P#1)
        PU L#3 (P#5)
      L2 L#2 (256KB) + L1d L#2 (32KB) + L1i L#2 (32KB) + Core L#2
        PU L#4 (P#2)
        PU L#5 (P#6)
      L2 L#3 (256KB) + L1d L#3 (32KB) + L1i L#3 (32KB) + Core L#3
        PU L#6 (P#3)
        PU L#7 (P#7)
  HostBridge
    PCIBridge
      PCI 01:00.0 (VGA)
    PCIBridge
      PCI 04:00.0 (Network)
[...]

PR #376, opened 2019-12-13
topology.c: fix undefined behavior of passing null to qsort()

senior7515:
stdlib.h:828 defines qsort as:

/* Sort NMEMB elements of BASE, of SIZE bytes each,
   using COMPAR to perform the comparisons.  */
extern void qsort (void *__base, size_t __nmemb, size_t __size,
		   __compar_fn_t __compar) __nonnull ((1, 4));

Explicitly marking arg1 and arg4 as nonnull which is caught by
the undefined behavior sanitizer -fsanitize=undefined
and running any program with environment variable set
UBSAN_OPTIONS="print_stacktrace=1"

This error was reported by clang-9.

Fixes: #373

### DIFF
--- a/src/topology.c
+++ b/src/topology.c
@@ -1393,7 +1393,9 @@ propagate_total_memory(hwloc_obj_t obj)
   /* By the way, sort the page_type array.
    * Cannot do it on insert since some backends (e.g. XML) add page_types after inserting the object.
    */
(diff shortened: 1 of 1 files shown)


bgoglin:
Please use 2.1, 1.11 is basically dead now. There's likely a similar warning there, but I am not sure I want to fix this almost-false-positive

PR #377, opened 2020-01-01
bsd: Make hwloc_get_sysctl able to return 64bit values

sthibaul:


sthibaul:
Mmm, after a merge update it doesn't start the CI??

PR #379, opened 2020-01-01
bsd: Make hwloc_get_sysctl able to return 64bit values

sthibaul:
Signed-off-by: Samuel Thibault <samuel.thibault@ens-lyon.org>


sthibaul:
Ah, ok, it was just waiting for an available worker. It's however dangerous that the PR list shows it as green even when the CI run hasn't proceeded yet.

PR #383, opened 2020-01-27
Execute hwloc_get_version.sh with sh

dtaskoff:
Long story short - I'm using **hwloc** on a system, on which I lose executable permissions, and since I have no control over this, I came to this solution, which, I thought, might be beneficial to someone else using **hwloc**.

### DIFF
--- a/contrib/dist/publish_doc
+++ b/contrib/dist/publish_doc
@@ -36,7 +36,7 @@ if ! test -f VERSION; then
   exit 1
 fi
 
-VERSION=$(./config/hwloc_get_version.sh ./VERSION)
+VERSION=$(sh config/hwloc_get_version.sh ./VERSION)
 if test x$VERSION = x; then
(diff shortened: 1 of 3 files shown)


contrib/dist/publish_doc
```diff
@@ -36,7 +36,7 @@ if ! test -f VERSION; then
   exit 1
 fi
 
-VERSION=$(./config/hwloc_get_version.sh ./VERSION)
+VERSION=$(sh /config/hwloc_get_version.sh ./VERSION)
```
jsquyres: I think you need to leave the `.` here, or ditch the `/`

dtaskoff: Sorry, I force-pushed it just before seeing your comments. I hope it's good now!

bgoglin:
Looks good to me, I merge (and backport to 2.1) after CI is done building/testing this.
Thanks a lot to both of you.

PR #397, opened 2020-04-09
Fix bash completion for zsh

rbellamy:
Currently the bash completion script breaks for zsh since `test ${BASH_VERSINFO[0]} -lt 4 && return` gives the error `/etc/bash_completion.d/hwloc-completion.bash:test:7: unknown condition: -lt` because `BASH_VERSINFO` is empty.

I'm not a shell wizard, so my fix is probably not idiomatic. I did make sure it works on Bash 3, 4, 5 and zsh.

### DIFF
--- a/contrib/completion/hwloc-completion.bash
+++ b/contrib/completion/hwloc-completion.bash
@@ -4,7 +4,7 @@
 #
 
 # bash < 4 doesn't support compopt
-test ${BASH_VERSINFO[0]} -lt 4 && return
+[[ -z "$ZSH_VERSION" ]] && [[ "${BASH_VERSINFO[0]:-0}" -lt 4 ]] && return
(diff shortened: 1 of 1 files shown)


bgoglin:
Hello
My zsh still complains after your change:
```
contrib/completion/hwloc-completion.bash:131: command not found: complete
```
Is there anything to enable in zsh for this to work?
Brice

bgoglin:
Ok somebody told me about bashcompinit, I am merging this and I will see if we need to document bashcompinit in the completion file itself.

bgoglin:
By the way, thanks!

rbellamy:
Sorry, I didn't see your question until just now! You're welcome.

About bashcompinit - it's a pretty standard install for folks using zsh who want to leverage the large ecosystem of bash completions.

PR #398, opened 2020-05-01
Add rsmi plugin for AMD GPU

miketxli:
Signed-off-by: Mike Li <Tianxinmike.Li@amd.com>

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -1019,6 +1019,30 @@ return clGetDeviceIDs(0, 0, 0, NULL, NULL);
     fi
     # don't add LIBS/CFLAGS/REQUIRES yet, depends on plugins
 
+    # RSMI support, rocm_smi64 is just library name and not related to 32/64 bits
+    hwloc_rsmi_happy=no
+    if test "x$enable_io" != xno && test "x$enable_rsmi" != "xno"; then
+      hwloc_rsmi_happy=yes
+      AC_CHECK_HEADERS([rocm_smi.h], [
+        AC_CHECK_LIB([rocm_smi64], [rsmi_init], [HWLOC_RSMI_LIBS="-lrocm_smi64"], [hwloc_rsmi_happy=no])
+        ], [hwloc_rsmi_happy=no])
+    fi
+    AC_SUBST(HWLOC_RSMI_LIBS)
+    # If we asked for rsmi support but couldn't deliver, fail
+    AS_IF([test "$enable_rsmi" = "yes" -a "$hwloc_rsmi_happy" = "no"],
+      [AC_MSG_WARN([Specified --enable-rsmi switch, but could not])
+      AC_MSG_WARN([find appropriate support])
+      AC_MSG_ERROR([Cannot continue])])
+    if test "x$hwloc_rsmi_happy" = "xyes"; then
+      AC_DEFINE([HWLOC_HAVE_RSMI], [1], [Define to 1 if you have the `RSMI' library.])
+      AC_SUBST([HWLOC_HAVE_RSMI], [1])
+      hwloc_components="$hwloc_components rsmi"
+      hwloc_rsmi_component_maybeplugin=1
+    else
+      AC_SUBST([HWLOC_HAVE_RSMI], [0])
+    fi
+    # don't add LIBS/CFLAGS/REQUIRES yet, depends on plugins
+
     # X11 support
     AC_PATH_XTRA
 
@@ -1291,6 +1315,10 @@ return clGetDeviceIDs(0, 0, 0, NULL, NULL);
           [HWLOC_LIBS="$HWLOC_LIBS $HWLOC_NVML_LIBS"
            HWLOC_CFLAGS="$HWLOC_CFLAGS $HWLOC_NVML_CFLAGS"
            HWLOC_REQUIRES="$HWLOC_NVML_REQUIRES $HWLOC_REQUIRES"])
+    AS_IF([test "$hwloc_rsmi_component" = "static"],
+          [HWLOC_LIBS="$HWLOC_LIBS $HWLOC_RSMI_LIBS"
+           HWLOC_CFLAGS="$HWLOC_CFLAGS $HWLOC_RSMI_CFLAGS"
+           HWLOC_REQUIRES="$HWLOC_RSMI_REQUIRES $HWLOC_REQUIRES"])
     AS_IF([test "$hwloc_gl_component" = "static"],
           [HWLOC_LIBS="$HWLOC_LIBS $HWLOC_GL_LIBS"
            HWLOC_CFLAGS="$HWLOC_CFLAGS $HWLOC_GL_CFLAGS"
@@ -1383,6 +1411,7 @@ AC_DEFUN([HWLOC_DO_AM_CONDITIONALS],[
         AM_CONDITIONAL([HWLOC_HAVE_PCIACCESS], [test "$hwloc_pciaccess_happy" = "yes"])
         AM_CONDITIONAL([HWLOC_HAVE_OPENCL], [test "$hwloc_opencl_happy" = "yes"])
         AM_CONDITIONAL([HWLOC_HAVE_NVML], [test "$hwloc_nvml_happy" = "yes"])
+        AM_CONDITIONAL([HWLOC_HAVE_RSMI], [test "$hwloc_rsmi_happy" = "yes"])
         AM_CONDITIONAL([HWLOC_HAVE_BUNZIPP], [test "x$BUNZIPP" != "xfalse"])
         AM_CONDITIONAL([HWLOC_HAVE_USER32], [test "x$hwloc_have_user32" = "xyes"])
 
@@ -1417,6 +1446,7 @@ AC_DEFUN([HWLOC_DO_AM_CONDITIONALS],[
         AM_CONDITIONAL([HWLOC_OPENCL_BUILD_STATIC], [test "x$hwloc_opencl_component" = "xstatic"])
         AM_CONDITIONAL([HWLOC_CUDA_BUILD_STATIC], [test "x$hwloc_cuda_component" = "xstatic"])
         AM_CONDITIONAL([HWLOC_NVML_BUILD_STATIC], [test "x$hwloc_nvml_component" = "xstatic"])
+        AM_CONDITIONAL([HWLOC_RSMI_BUILD_STATIC], [test "x$hwloc_rsmi_component" = "xstatic"])
         AM_CONDITIONAL([HWLOC_GL_BUILD_STATIC], [test "x$hwloc_gl_component" = "xstatic"])
         AM_CONDITIONAL([HWLOC_XML_LIBXML_BUILD_STATIC], [test "x$hwloc_xml_libxml_component" = "xstatic"])
 
--- a/config/hwloc_internal.m4
+++ b/config/hwloc_internal.m4
@@ -67,7 +67,7 @@ AC_DEFUN([HWLOC_DEFINE_ARGS],[
     # I/O?
     AC_ARG_ENABLE([io],
                   AS_HELP_STRING([--disable-io],
-                                 [Disable I/O discovery build entirely (PCI, LinuxIO, CUDA, OpenCL, NVML, GL) instead of only disabling it at runtime by default]))
+                                 [Disable I/O discovery build entirely (PCI, LinuxIO, CUDA, OpenCL, NVML, RSMI, GL) instead of only disabling it at runtime by default]))
 
     # PCI?
     AC_ARG_ENABLE([pci],
@@ -92,6 +92,11 @@ AC_DEFUN([HWLOC_DEFINE_ARGS],[
     AC_ARG_ENABLE([nvml],
                   AS_HELP_STRING([--disable-nvml],
                                  [Disable the NVML device discovery build (instead of only disabling NVML at runtime by default)]))
+
+    # RSMI?
+    AC_ARG_ENABLE([rsmi],
+                  AS_HELP_STRING([--disable-rsmi],
+                                 [Disable the ROCm SMI device discovery]))
 
     # GL/Display
     AC_ARG_ENABLE([gl],
@@ -498,6 +503,7 @@ int foo(void) {
 	hwloc_config_prefix[tests/hwloc/ports/topology-opencl.c]:hwloc_config_prefix[hwloc/topology-opencl.c]
 	hwloc_config_prefix[tests/hwloc/ports/topology-cuda.c]:hwloc_config_prefix[hwloc/topology-cuda.c]
 	hwloc_config_prefix[tests/hwloc/ports/topology-nvml.c]:hwloc_config_prefix[hwloc/topology-nvml.c]
+	hwloc_config_prefix[tests/hwloc/ports/topology-rsmi.c]:hwloc_config_prefix[hwloc/topology-rsmi.c]
 	hwloc_config_prefix[tests/hwloc/ports/topology-gl.c]:hwloc_config_prefix[hwloc/topology-gl.c]
 	hwloc_config_prefix[tests/hwloc/ports/lstopo-windows.c]:hwloc_config_prefix[utils/lstopo/lstopo-windows.c])
     ])
--- a/hwloc/topology-rsmi.c
+++ b/hwloc/topology-rsmi.c
@@ -1,0 +1,243 @@
+/*
+ * Copyright © 2012-2020 Inria.  All rights reserved.
+ * Copyright (c) 2020, Advanced Micro Devices, Inc. All rights reserved.
+ * Written by Advanced Micro Devices,
+ * See COPYING in top-level directory.
+ */
+
+#include "private/autogen/config.h"
+#include "hwloc.h"
+#include "hwloc/plugins.h"
+
+/* private headers allowed for convenience because this plugin is built within hwloc */
+#include "private/misc.h"
+#include "private/debug.h"
+
+#include <rocm_smi.h>
+
+/*
+ * Get the name of the GPU
+ *
+ * dv_ind		(IN) The device index
+ * device_name	(OUT) Name of GPU devices
+ * size			(OUT) Size of the name
+ */
+static int get_device_name(uint32_t dv_ind, char *device_name, unsigned int size)
+{
+  const char *status_string;
+  rsmi_status_t rsmi_rc = rsmi_dev_name_get(dv_ind, device_name, size);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    rsmi_rc = rsmi_status_string(rsmi_rc, &status_string);
+    fprintf(stderr, "RSMI: GPU(%d): Failed to get name: %s\n", dv_ind, status_string);
+    return -1;
+  }
+  return 0;
+}
+
+/*
+ * Get the PCI Info of the GPU
+ *
+ * dv_ind  (IN) The device index
+ * bdfid   (OUT) PCI Info of GPU devices
+ */
+static int get_device_pci_info(uint32_t dv_ind, uint64_t *bdfid)
+{
+  const char *status_string;
+  rsmi_status_t rsmi_rc = rsmi_dev_pci_id_get(dv_ind, bdfid);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    rsmi_rc = rsmi_status_string(rsmi_rc, &status_string);
+    fprintf(stderr, "RSMI: GPU(%d): Failed to get PCI Info: %s\n", dv_ind, status_string);
+    return -1;
+  }
+  return 0;
+}
+
+/*
+ * Get the PCI link speed of the GPU
+ *
+ * dv_ind    (IN) The device index
+ * linkspeed (OUT) PCI link speed of GPU devices
+ */
+static int get_device_pci_linkspeed(uint32_t dv_ind, float *linkspeed)
+{
+  rsmi_pcie_bandwidth_t bandwidth;
+  uint64_t lanespeed_raw; // T/s
+  uint64_t lanespeed; // (bits/s)
+  uint32_t lanes;
+  rsmi_status_t rsmi_rc = rsmi_dev_pci_bandwidth_get(dv_ind, &bandwidth);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    return -1;
+  }
+  lanespeed_raw = bandwidth.transfer_rate.frequency[bandwidth.transfer_rate.current]; // T/s
+  lanespeed = lanespeed_raw <= 5000000000 ? (lanespeed_raw * 8)/10 : (lanespeed_raw * 128)/130; // bits/s
+  lanes = bandwidth.lanes[bandwidth.transfer_rate.current];
+  *linkspeed = (lanespeed * lanes) / 8000000000; // GB/s
+  return 0;
+}
+
+/*
+ * Get the Unique ID of the GPU
+ *
+ * dv_ind  (IN) The device index
+ * buffer  (OUT) Unique ID of GPU devices
+ */
+static int get_device_unique_id(uint32_t dv_ind, char *buffer)
+{
+  uint64_t id;
+  rsmi_status_t rsmi_rc = rsmi_dev_unique_id_get(dv_ind, &id);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    return -1;
+  }
+  sprintf(buffer, "%lx", id);
+  return 0;
+}
+
+/*
+ * Get the serial number of the GPU
+ *
+ * dv_ind  (IN) The device index
+ * serial  (OUT) Serial number of GPU devices
+ * size    (IN) Length of the caller provided buffer
+ */
+static int get_device_serial_number(uint32_t dv_ind, char *serial, unsigned int size)
+{
+  rsmi_status_t rsmi_rc = rsmi_dev_serial_number_get(dv_ind, serial, size);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    return -1;
+  }
+  return 0;
+}
+
+static int
+hwloc_rsmi_discover(struct hwloc_backend *backend, struct hwloc_disc_status *dstatus)
+{
+  /*
+   * This backend uses the underlying OS.
+   * However we don't enforce topology->is_thissystem so that
+   * we may still force use this backend when debugging with !thissystem.
+   */
+
+  struct hwloc_topology *topology = backend->topology;
+  enum hwloc_type_filter_e filter;
+  rsmi_status_t ret;
+  unsigned nb, i;
+
+  assert(dstatus->phase == HWLOC_DISC_PHASE_IO);
+
+  hwloc_topology_get_type_filter(topology, HWLOC_OBJ_OS_DEVICE, &filter);
+  if (filter == HWLOC_TYPE_FILTER_KEEP_NONE)
+    return 0;
+
+  rsmi_init(0);
+
+  ret = rsmi_num_monitor_devices(&nb);
+  if (RSMI_STATUS_SUCCESS != ret || !nb) {
+    if (RSMI_STATUS_SUCCESS != ret)
+      fprintf(stderr, "RSMI: Failed to get device count\n");
+    rsmi_shut_down();
+    return 0;
+  }
+
+  for (i=0; i<nb; i++) {
+    uint64_t bdfid = 0;
+    unsigned domain, device, bus, func;
+    hwloc_obj_t osdev, parent;
+    char buffer[64];
+
+    osdev = hwloc_alloc_setup_object(topology, HWLOC_OBJ_OS_DEVICE, HWLOC_UNKNOWN_INDEX);
+    snprintf(buffer, sizeof(buffer), "rsmi%u", i);
+    osdev->name = strdup(buffer);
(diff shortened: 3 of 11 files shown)


miketxli:
Install rocm-smi library pkg following
https://rocm-documentation.readthedocs.io/en/latest/Installation_Guide/Installation-Guide.html

Then
cd hwloc
./autogen.sh
./configure --enable-rsmi --enable-plugins='rsmi' CPPFLAGS=-I/opt/rocm/rocm_smi/include/rocm_smi LDFLAGS=-L/opt/rocm/rocm_smi/lib --enable-shared

config/hwloc.m4
```diff
@@ -1019,6 +1019,39 @@ return clGetDeviceIDs(0, 0, 0, NULL, NULL);
...
+    fi
+    if test "x$hwloc_rsmi_happy" = "xyes"; then
+      tmp_save_CFLAGS="$CFLAGS"
+      CFLAGS="$CFLAGS $HWLOC_RSMI_CFLAGS"
+      tmp_save_LIBS="$LIBS"
+      LIBS="$LIBS $HWLOC_RSMI_LIBS"
+      # AC_CHECK_DECLS([nvmlDeviceGetMaxPcieLinkGeneration],,[:],[[#include <nvml.h>]])
+      # CFLAGS="$tmp_save_CFLAGS"
+      # LIBS="$tmp_save_LIBS"
+    fi
```
bgoglin, line 1038: If you don't have any specific function to check (something that was added after the first RSMI release and that you want to put under #ifdef in hwloc), remove this entire "if/fi" block.

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+      uint64_t function : 3;
+      uint64_t device : 5;
+      uint64_t bus : 8;
+      uint64_t reserved : 16;
+      uint64_t domain : 32;
+#endif
+    };
+    uint64_t bdfid;
+  };
+} rsmiPciInfo_t;
```
bgoglin, line 41: Looks like we need to add some configury to define HWLOC_BIGENDIAN?
Also isn't this structure defined in RSMI headers? I'd be better not to duplicate it in hwloc.

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+ * size			(OUT) Size of the name
+ */
+static void get_device_name(uint32_t dv_ind, char *device_name, unsigned int size)
+{
+  const char *status_string;
+  rsmi_status_t rsmi_rc = rsmi_dev_name_get(dv_ind, device_name, size);
+
+  if (rsmi_rc != RSMI_STATUS_SUCCESS) {
+    rsmi_rc = rsmi_status_string(rsmi_rc, &status_string);
+    fprintf(stderr, "Failed to get name of the GPU: %s\n", status_string);
```
bgoglin, line 57: Make sure these error messages only occur under catastrophic circumstance sicne we don't want the hwloc library to pollute the stderr of programs that don't even know there are using hwloc.
Also add a ref to RSMI and/or rsmi_dev_name() in the message.
Same for all other fprintf(stderr...) below.

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+    get_device_pci_info(i, &pci);
+    parent = hwloc_pci_find_parent_by_busid(topology, pci.domain, pci.bus, pci.device, 0);
+    get_device_pci_linkspeed(i, &(parent->attr->pcidev.linkspeed));
+    if (!parent)
+      parent = hwloc_get_root_obj(topology);
+
+    hwloc_insert_object_by_parent(topology, parent, osdev);
+  }
+
+  rsmi_shut_down();
```
bgoglin, line 213: What happens if the program does its own rsmi_init() either earlier or before? do you have proper reference counting inside RSMI to make sure this rsmi_shut_down() in hwloc shut down RSMI for the entire process?

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+  backend->discover = hwloc_rsmi_discover;
+  return backend;
+}
+
+static struct hwloc_disc_component hwloc_rsmi_disc_component = {
+  "rsmi",
+  HWLOC_DISC_PHASE_IO,
+  HWLOC_DISC_PHASE_GLOBAL,
+  hwloc_rsmi_component_instantiate,
+  5, /* after pci, and after cuda since likely less useful */
```
bgoglin, line 239: Do you really want this ref to CUDA being more useful ? :)
If you want to compare to other IO backends, opencl might be the relevant one since it may also add osdevices for the same GPUs. Do you care whether rsmi is before or after OpenCL? If yes, adjust the priority above or below OpenCL.
Otherwise, keep the same priority as CUDA etc.

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+hwloc_rsmi_component_init(unsigned long flags)
+{
+  if (flags)
+    return -1;
+  if (hwloc_plugin_check_namespace("rsmi", "hwloc_backend_alloc") < 0)
+    return -1;
+  if (!dlopen("librocm_smi64.so", RTLD_NOW | RTLD_GLOBAL)) {
+    fprintf(stderr, "RSMI configured, but wasn't found.\n");
+    return -1;
+  }
```
bgoglin, line 254: Why do you need this dlopen? The rsmi hwloc plugin won't be loaded unless this librocm_smi64.so is available since it will depend on it.

config/hwloc.m4
```diff
@@ -1019,6 +1019,39 @@ return clGetDeviceIDs(0, 0, 0, NULL, NULL);
     fi
     # don't add LIBS/CFLAGS/REQUIRES yet, depends on plugins
 
+    # RSMI support
+    hwloc_rsmi_happy=no
+    if test "x$enable_io" != xno && test "x$enable_rsmi" != "xno"; then
+      hwloc_rsmi_happy=yes
+      AC_CHECK_HEADERS([rocm_smi.h], [
+        AC_CHECK_LIB([rocm_smi64], [rsmi_init], [HWLOC_RSMI_LIBS="-lrocm_smi64"], [hwloc_rsmi_happy=no])
+        ], [hwloc_rsmi_happy=no])
```
bgoglin, line 1028: Is there a librocm_smi32.so for 32bits systems?
Add a comment so that we know why we're only looking for the 64bits lib without checking if we're on a 64bit machine.

include/hwloc/rsmi.h
```diff
@@ -0,0 +1,214 @@
...
+      uint64_t function : 3;
+      uint64_t device : 5;
+      uint64_t bus : 8;
+      uint64_t reserved : 16;
+      uint64_t domain : 32;
+#endif
+    };
+    uint64_t bdfid;
+  };
+} rsmiPciInfo_t;
```
bgoglin, line 64: If we keep this structure in hwloc as discussed above, remove it from topology-rsmi.c and #include hwloc/rsmi.h in topology-rsmi.c
Also we'll need HWLOC_BIGENDIAN to be exported to the *public* autogen/config.h since application will need the #define when they'll use hwloc/rsmi.h.
(I can take care of HWLOC_BIGENDIAN).

include/hwloc/rsmi.h
```diff
@@ -0,0 +1,214 @@
...
+
+    info = hwloc_obj_get_info_by_name(osdev, "AMDUUID");
+    if (info && !strcmp(info, uuid))
+      return osdev;
+  }
+
+  return NULL;
+}
+
+/** @} */
```
bgoglin, line 206: There was some idea of moving some of these helpers to somewhere else. If it's not a priority for you, I'd rather keep them in a separate commit so that we can merge the core backend and discuss helpers laters.

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,268 @@
...
+    char buffer[64];
+
+    osdev = hwloc_alloc_setup_object(topology, HWLOC_OBJ_OS_DEVICE, HWLOC_UNKNOWN_INDEX);
+    snprintf(buffer, sizeof(buffer), "rsmi%u", i);
+    osdev->name = strdup(buffer);
+    osdev->depth = HWLOC_TYPE_DEPTH_UNKNOWN;
+    osdev->attr->osdev.type = HWLOC_OBJ_OSDEV_GPU;
+
+    hwloc_obj_add_info(osdev, "Backend", "RSMI");
+    hwloc_obj_add_info(osdev, "GPUVendor", "Advanced Micro Devices");
```
bgoglin, line 187: Would "AMD" be acceptable here? This kind of string is usually printed in verbose message, so having something short and easier to recognize wouldn't hurt.

include/hwloc/rsmi.h
```diff
@@ -0,0 +1,214 @@
...
+
+    if (pcidev
+      && pcidev->type == HWLOC_OBJ_PCI_DEVICE
+      && pcidev->attr->pcidev.domain == pci.domain
+      && pcidev->attr->pcidev.bus == pci.bus
+      && pcidev->attr->pcidev.dev == pci.device
+      && pcidev->attr->pcidev.func == pci.function)
+      return osdev;
+
+    info = hwloc_obj_get_info_by_name(osdev, "AMDUUID");
```
bgoglin, line 198: I don't know if NVIDIAUUID was a good idea. Maybe just use UUID here?

tests/hwloc/ports/include/rsmi/rsmi.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+#ifndef HWLOC_PORT_RSMI_RSMI_H
+#define HWLOC_PORT_RSMI_RSMI_H
+
+/* we need to replace any RSMI-related #define that configure may have put in private/autogen/config.h */
+#ifndef HWLOC_CONFIGURE_H
+#error rsmi.h must be included after private/autogen/config.h
+#endif
+
```
bgoglin, line 15: This file should be named exactly as the official headers used in topology-rsmi.c so I guess you should replace rsmi/rsmi.h with rocm_smi.h. That's likely the reason why our CI failed.

miketxli:
Thanks, Brice

Let me go through the comments and come back to you.

Thanks

Mike

config/hwloc.m4
miketxli: will remove it.

hwloc/topology-rsmi.c
miketxli: This structure isn't defined in RSMI headers. We can remove HWLOC_BIGENDIAN, since it's not used.

hwloc/topology-rsmi.c
miketxli: OK

hwloc/topology-rsmi.c
bgoglin: Oh I see, only the bdfid part of union seems to be passed to rsmi_dev_pci_id_get(), the struct is here for decoding only. Does RSMI work on some big endian architectures?

hwloc/topology-rsmi.c
miketxli: The first time init() is called, the internal data structures are initialized, and the first time shutdown is called they are freed.

miketxli: There is no reference counting inside RSMI right now, can be added later.

hwloc/topology-rsmi.c
miketxli: Yes, that's for decoding only. I have ashed some regarding big endian architectures.

hwloc/topology-rsmi.c
bgoglin: It *must* be added as soon as possible. Missing support for multiple init/shutdown is very annoying for developers. Right now, it means hwloc just cannot call shutdown at all: if an application (e.g using hwloc/rsmi.h) calls rsmi_init() before hwloc_topology_load(), hwloc would shutdown rsmi at the end of that hwloc_topology_load().

Not calling shutdown means valgrind will complain. Hence we'll end up adding yet another environment variable sayign "OK hwloc, you can shutdown rsmi, I know what I am doing" so that our regression tests don't fail.

miketxli: Let me talk to the developer.

tests/hwloc/ports/include/rsmi/rsmi.h
bgoglin: Fix what I said above: the file should be in include/rsmi/rocm_smi.h (headers used to simulate rsmi build without rsmi are in include/rsmi/)
You may test this part with make check -C tests/hwloc/ports

hwloc/topology-rsmi.c
bgoglin: Given that the rsmi doc explicitly says:
```
 *      BDFID = ((DOMAIN & 0xffffffff) << 32) | ((BUS & 0xff) << 8) |
 *                                   ((DEVICE & 0x1f) <<3 ) | (FUNCTION & 0x7)
```
I would just remove the structure entirely and decode the uint64_t manually, it won't depend on big/little endian.

hwloc/topology-rsmi.c
bgoglin: I put all details in a github issue against rsmi https://github.com/RadeonOpenCompute/rocm_smi_lib/issues/62

bgoglin:
Should we also call rsmi_dev_memory_total_get() to annotate the hwloc object with information like we do for opencl and cuda?
OpenCLGlobalMemorySize=3087360
OpenCLComputeUnits=68
CUDAGlobalMemorySize=16548352
CUDAL2CacheSize=6144
CUDAMultiProcessors=80
CUDACoresPerMP=64
CUDASharedMemorySizePerMP=48

hwloc/topology-rsmi.c
miketxli: Ok

hwloc/topology-rsmi.c
miketxli: Will put it before OpenCL

include/hwloc/rsmi.h
miketxli: Will use decode

include/hwloc/rsmi.h
miketxli: agree

hwloc/topology-rsmi.c
miketxli: OK

include/hwloc/rsmi.h
miketxli: OK

tests/hwloc/ports/include/rsmi/rsmi.h
miketxli: You mean it should be in tests/hwloc/ports/include/rsmi directory, but change file name to rocm_smi.h same as in topology-rsmi.c?

bgoglin: Yes!

hwloc/topology-rsmi.c
miketxli: I will forward it to the developer for rocm_smi library. Like I said at beginning, this is just phase 1 for hwloc plugin. It will be added in phase 2.

miketxli:
We can add those in phase 2.

bgoglin:
OK
Feel free to add some words in doc/hwloc.dox (when we list dependencies, info attributes, OS device names, and interoperability helpers). Look for NVIDIA and add RSMI things nearby :)
I still haven't figured out whether it's better to name your info attribute AMDFoo or just a generic Foo.

config/hwloc.m4
miketxli: There is no  librocm_smi32.so, they call library  librocm_smi64.so

hwloc/topology-rsmi.c
miketxli: Let see hwloc is build in another machine. But rocm_smi isn't installed in the machine hwloc is running, we stop here instead crash when with init.

bgoglin: If topology-rsmi.c is built statically inside libhwloc (not as plugin), libhwloc.so will depend on librocm_smi64.so explicitly, hence libhwloc.so will fail to load, and the program will fail to start because of the missing librocm64_smi.so
If built as a plugin, libhwloc.so will work fine, the program will start. When hwloc will load plugins, libhwloc_rsmi.so will fail to load because it needs librocm_smi64.so. That's actually the whole point of plugins: build everything but don't require all dependencies to be available at runtime, only those available will have their hwloc plugins loaded.
It'll work fine without dlopen, we already do that for multiple other dependencies.

config/hwloc.m4
bgoglin: Indeed the CMakeLists.txt only enables 64bits by default, and manual changes to switch to 32bits don't touch the library name. Please put a comment about this :)

hwloc/topology-rsmi.c
miketxli: Ok

config/hwloc.m4
miketxli: will add a conmments.

miketxli:
I will go ahead and make update based on what we discussed. Are you OK with that?

bgoglin:
Yes sure, go ahead. You may even force push to replace your previous commit with a better one.

miketxli:
I'm multitasking right now. It will take some time for me to come back.
Thanks

Mike

miketxli:
Sorry about delay. I'm working on this task now. The good news is that they have added a ref count for rocm-smi library. 
I have an issue after merged with latest code. 
It used to show plugin (rsmi), now shows dlopen(rsmi) after configuration. When I run lstopo, I got segmentation fault. It has something to do with recent change for AIX. I'm debug it. Maybe you can point me to right direction.

Thanks

Mike

miketxli:
I fixed the issue by disable dlopen.

Thanks

Mike

bgoglin:
Can you clarify what you meant by "recent change for AIX"?
I review the code next week, thanks.

bgoglin:
CI failed, looks like you need to add rsmi_status_string to tests/hwloc/ports/include/rsmi/rocm_smi.h
```
In file included from topology-rsmi.c:17:
/builds/workspace/basic_PR-398/hwloc-PR-398-20200522.2236.git60689bb/tests/hwloc/ports/include/rsmi/rocm_smi.h:37: error: ‘RSMI_MAX_NUM_FREQUENCIES’ undeclared here (not in a function)
topology-rsmi.c: In function ‘get_device_name’:
topology-rsmi.c:32: warning: implicit declaration of function ‘rsmi_status_string’
```

miketxli:
It's related to the following check in 
commit 9c02aa1a76d2422cc13d38ddaddfcc23087fe8aa
Author: Brice Goglin <Brice.Goglin@inria.fr>
Date:   Mon Feb 24 17:11:46 2020 +0100

    configure: use dlopen instead of ltdl by default
    
I have to disable dlopen to use ltdl.

miketxli:
I fix the compile issue.

bgoglin:
Which Linux distribution do you use? I'll see if I can test the dlopen code on a similar one here.

miketxli:
Ubuntu 18.04, I have to do the following:
./configure --disable-plugin-dlopen --enable-rsmi --enable-plugins='rsmi' CPPFLAGS="-I/opt/rocm/rocm_smi/include/rocm_smi -I/opt/rocm/rocm_smi/include" LDFLAGS=-L/opt/rocm/rocm_smi/lib --enable-shared
to disable dlopen

bgoglin:
We're trying to reproduce the dlopen issue here, and installing RSMI to test the code. In the meantime, here are a couple minor changes:
* Remove #include dlfcn.h
* Change all small get_device_foo() helper function to return 0/-1 on success error, or set the buffer to sane default values on error (e.g. set the linkspeed to 0).
* Change back UUID to AMDUUID since you still have AMDSerial, and we cannot easily change NVIDIA info key names anymore. Sorry for changing my mind here.

miketxli:
Have updated base on the comments.
thanks
Mike

bgoglin:
Thanks. Is the kernel driver in the upstream Linux kernel? Which minimal kernel version do we need? We couldn't reproduce the dlopen crash so far, but rsmi_init() fails because of the missing kernel driver :)

config/hwloc_internal.m4
```diff
@@ -93,6 +93,11 @@ AC_DEFUN([HWLOC_DEFINE_ARGS],[
                   AS_HELP_STRING([--disable-nvml],
                                  [Disable the NVML device discovery build (instead of only disabling NVML at runtime by default)]))
 
+    # RSMI?
+    AC_ARG_ENABLE([rsmi],
+                  AS_HELP_STRING([--disable-rsmi],
+                                 [Disable the RSMI device discovery]))
+
```
bgoglin, line 100: Can you write something more clear such "ROCm SMI" in this documentation line?

tests/hwloc/ports/include/rsmi/rocm_smi.h
```diff
@@ -0,0 +1,69 @@
...
+ * See COPYING in top-level directory.
+ */
+
+#ifndef HWLOC_PORT_RSMI_RSMI_H
+#define HWLOC_PORT_RSMI_RSMI_H
+
+/* we need to replace any RSMI-related #define that configure may have put in private/autogen/config.h */
+#ifndef HWLOC_CONFIGURE_H
+#error rsmi.h must be included after private/autogen/config.h
+#endif
```
bgoglin, line 14: This #error should say rocm_smi.h instead of rsmi.h

config/hwloc_internal.m4
miketxli: done

tests/hwloc/ports/include/rsmi/rocm_smi.h
miketxli: Ok, good finding

miketxli:
We should be able to use any version of driver, but lets try with latest up steamed.

bgoglin:
I am preparing some documentation updates. They raised 2 questions:
* did you consider using OSDEV subtype COPROC instead of GPU? The difference is small but I assume ROCm is not only about graphics but also about computing. CUDA uses COPROC. NVML uses GPU because it's really for management, and they create separate OS devices. I'd like to cleanup those subtypes in the future but it's not feasible for now
* I found the Debian packages name 'librocm-smi-dev" (Debian will ship official ROCm packages soon) but I am not sure about RPM package names. I found rocm_smi.h in "rocm-smi-lib64" in the AMD repo but this name looks a bit awkward to me. Do you know if that's what RHEL might ship in the future?

miketxli:
For the first question, I would use GPU. RSMI is used to manager GPU, similar to NVML. The compute API is provided by HIP, similar to CUDA.
For the second question, I will ask the developer and come back to you.

tests/hwloc/ports/include/rsmi/rocm_smi.h
bgoglin: typo in rsmi_smi.h :)

bgoglin:
Here's the doc update I was talking about. Feel free to include them in your commit, or I 'll push them after merging your PR.
https://github.com/bgoglin/hwloc/commit/6c501ad8f7d53a856ba62e1c15f242026cec9ded

miketxli:
Updated based on your last comments.
For 1.st question, I have asked some high level architects and is waiting for comments.
For 2nd question,  it should be the same for both deb and rpm.

Thanks

Mike

bgoglin:
I am finally testing this on a real machine with a radeon fiji nano. The plugin works fine with dlopen here on Centos7 with rocm-smi-lib64.x86_64 package version 2.2.0.3.rocm_rel_3.3_19_a482394-1

get_uuid fails (I added a status string printf but it's NULL).
get_serial returns an empty string.
name, linkspeed and BDF look correct.

Can you change the code to do these?
* not set AMDSerial and AMDUUID when you get an error (instead of setting to empty string)
* fix the if for PCI BDF in the main function, it should check for ==0 instead of -1
* not set the BFD to 0 when you get an error. just set parent to hwloc_get_root_obj(topology); too, in that case.
* by the way, the status string variable may move to inside the ifs
* why do you printf in some cases and not others? people running apps over MPI over hwloc might wonder why they get these error even if they don't care about hwloc. The core hwloc has a way to hide such messages when HWLOC_HIDE_ERRORS=1 in the env (but it's not available to plugins, so we'd to duplicate getenv here).

I can take care of these changes now that I can test things. But I'd need your help about why rocm_smi return empty serial and fails to return the uuid.

miketxli:
get_uuid fails (I added a status string printf but it's NULL).
get_serial returns an empty string.
uuid and serial are supported only by newer GPU's, Fiji doesn't support it.

I made fix for other things. You can change it if it's minor.

Thanks

Mike

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,243 @@
...
+    if (get_device_pci_info(i, &bdfid) == 0) {
+      domain = (bdfid>>32) & 0xffffffff;
+      bus = ((bdfid & 0xffff)>>8) & 0xff;
+      device = ((bdfid & 0xff)>>3) & 0x1f;
+      func = bdfid & 0x7;
+      parent = hwloc_pci_find_parent_by_busid(topology, domain, bus, device, func);
+      if (parent) {
+        if (get_device_pci_linkspeed(i, &(parent->attr->pcidev.linkspeed)) == -1)
+          parent->attr->pcidev.linkspeed = 0;
+      } else {
```
bgoglin, line 184: add "&& parent->type == HWLOC_OBJ_PCI_DEVICE" in "if (parent)".

bgoglin:
I think I am ready to merge this beside the small change I just put in the review.
Is it ready to merge on your side too? After this, I'll add envvars HWLOC_HIDE_ERRORS=1 (to optionally silence error messages) and HWLOC_RSMI_NO_SHUTDOWN=1 (to optionally disable rsmi_shutdown() if it breaks things when running a librsmi without refcounting).

hwloc/topology-rsmi.c
```diff
@@ -0,0 +1,243 @@
...
+
+    buffer[0] = '\0';
+    if (get_device_name(i, buffer, sizeof(buffer)) == -1)
+      buffer[0] = '\0';
+    hwloc_obj_add_info(osdev, "GPUModel", buffer);
+
+    buffer[0] = '\0';
+    if (get_device_serial_number(i, buffer, sizeof(buffer)) == 0)
+      hwloc_obj_add_info(osdev, "AMDSerial", buffer);
+
```
bgoglin, line 169: if buffer can be left empty, add "&& buffer[0]" at the end of the if() above

miketxli:
Added type check. If you ready, you can merge. I don't have more to change.

Thanks

Mike

bgoglin:
I just put yet another comment in review. Otherwise testing looks good here.

hwloc/topology-rsmi.c
miketxli: buffer can contain some data, even so call failed.

bgoglin: What I am getting on Fiji is not a failure but an empty buffer, so I wanted to ignore this empty AMDSerial.

miketxli: I didn't that. Ok, I will add check.

Thanks

Mike

bgoglin:
@miketxli Now that we merged this, I seem to remember you had a hwloc/rsmi.h and a test case in the first submission? Can you push them back to a branch so that we look at them now?

miketxli:
Ok, let me take a look. Can you tell me how to compile and test that?

bgoglin:
Tarballs of git master are updated here after each commit/push https://ci.inria.fr/hwloc/job/basic/job/master/

bgoglin:
@miketxli Hello, 2 questions.

hwloc/rsmi doesn't build over a manually-built rocm_smi_lib from github master because rocm_smi.h now includes rocm_smi/kfd_ioctl.h. It looks like to include rocm_smi/rocm_smi.h instead of rocm_smi.h (and remove the ending "rocm_smi/" from the include path). Can you confirm this?

I tried on a machine with 2 tahiti GPUs, rocm_smi doesn't seem to report any GPU. Are these cards too old?

miketxli:
For 1.st question, yes, they have added one more header file with another path. I have made it work by doing this: 
./configure --disable-plugin-dlopen --enable-rsmi --enable-plugins='rsmi' CPPFLAGS="-I/opt/rocm/rocm_smi/include/rocm_smi -I/opt/rocm/rocm_smi/include" LDFLAGS=-L/opt/rocm/rocm_smi/lib --enable-shared

For 2nd question, I haven't tested with any GPU older than Fiji. I will get the information for you.

Thanks

Mike

miketxli:
For 1.st question, they will take out the extra include later.
For 2nd question, the tahiti GPUs may not be supported by default through the kernel driver. How many AMD cards do you have? If you need newer cards, AMD can send you some sample cards.

bgoglin:
1) Given that rocm_smi_lib uses rocm_smi/rocm_smi.h everywhere, including in its own example, I changed the code to do the same. So now we configure with rocm_smi/include instead of rocm_smi/include/rocm_smi
2) We only have 2 very old tahiti (bought in 2012) and 2 fiji nano xt (sent by AMD a couple years ago when we started looking at HSA). If AMD can send some newer cards, it would be great. We'll need cards whose cooling is compatible with rack servers.

bgoglin:
Hello @miketxli Don't you have your original submission somewhere where you included hwloc/rsmi.h and tests/hwloc/rsmi.c? I will be able to test those now.

miketxli:
I have created pull request 404, have tested under Ubuntu 18.04

PR #401, opened 2020-05-26
configure.ac: remove "tar-ustar"

jsquyres:
Commit ba3ba61e88122d8aada992371eb12d23a0c06d34 introduced "tar-ustar" into the AM_INIT_AUTOMAKE line, citing the need for filenames longer than 99 characters in the dist tarball.

On MacOS 10.15.x (Catalina), the default "tar" option seems to allow for filenames longer than 99 characters (e.g., `hwloc-2.3.0a1/doc/doxygen-doc/man/man3/hwloc_topology_diff_obj_attr_u_hwloc_topology_diff_obj_attr_generic_s.3`, which is 111 characters long).  Hence, the need for `tar-ustar` does not seem to exist any longer.

Indeed, we have actually run into a developer who has run into a documented problem with the ustar format: it doesn't support UIDs greater than 21 bits in length, causing `make dist` to fail (and his UID is greater than 21 bits).  This developer was working in another Open MPI project (not hwloc), but we thought we'd bring this learned knowledge over to hwloc as well.

After some testing, it looks like not specifying a `tar-*` option to `AM_INIT_AUTOMAKE` causes Automake to not pass the `--format` option to `tar`, which -- at least with GNU `tar` -- defaults to `--format=gnu`.  This format has wide portability -- e.g., it works on old versions of NetBSD where tarballs created with the `tar-pax` Automake option do not work.

See https://github.com/openpmix/prrte/pull/579#issuecomment-632849426 for a little more detail.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,7 +37,7 @@ EOF
 AC_CANONICAL_TARGET
 
 # Init automake
-AM_INIT_AUTOMAKE([1.11 dist-bzip2 subdir-objects foreign tar-ustar parallel-tests -Wall -Werror])
+AM_INIT_AUTOMAKE([1.11 dist-bzip2 subdir-objects foreign parallel-tests -Wall -Werror])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 AC_LANG([C])


bgoglin:
Good, I saved the similar OMPI commit and wanted to ping you about it. Unfortunately, looks like we have troubles with long filenames causing the build to fail (not only in manpages).
```
tar: hwloc-PR-401-20200526.1649.gitddf75fd/doc/doxygen-doc/man
/man3/hwloc_topology_diff_u_hwloc_topology_diff_generic_s.3: file name is too long (max 99); not dumped
```

bgoglin:
The failure occured on a Centos6 box (where we build the tarball before passing it to other slaves for make check) with the AM 1.15 LT 246 AC 269 tar 1.23.

jsquyres:
CentOS 6: that makes sense -- it's an older version of `tar`.

That being said, would it be possible to move the create-tarballs machine to be a more modern distro?  FWIW, I see:

| Distro | GNU Tar version | Works / doesn't work |
|------|-----|-----|
| RHEL 6 / Centos 6 | 1.23 | Doesn't work |
| RHEL 7 / Centos 7 | 1.26 | Doesn't work |
| RHEL 8 / Centos 8 | 1.30 | Works |
| Ubuntu 18 | 1.29 | Works |
| Ubuntu 20 | 1.30 | Works |
| MacOS 10.15 (Catalina) | bsdtar 3.3.2 | Works |

I did test making a tarball on MacOS and untarring it on RHEL 6 -- that seemed to work fine.  So it's just the *creation* of tarballs with long filenames that is the problem.

bgoglin:
I'll build a more recent tar on those machines.

bgoglin:
Don't know why it still fails. It has tar 1.32 now, and I also changed the config so that it takes recent autotools even on the non-stable branches. The log is https://ci.inria.fr/hwloc/job/basic/view/change-requests/job/PR-401/3/consoleFull

jsquyres:
You might need to `./configure TAR=/path/to/tar` (or maybe `AM_TAR`)...?

bgoglin:
I added TAR=/path/to/patgh in the env, the full path now appears in the log, it's using the right tar but still failing because of long filenames. Am I supposed to see some new options on the tar command line ?
```
tardir=hwloc-PR-401-20200527.0045.gitddf75fd && ${TAR-tar} chof - "$tardir" | BZIP2=${BZIP2--9} bzip2 -c >hwloc-PR-401-20200527.0045.gitddf75fd.tar.bz2
/builds/local/bin/tar: hwloc-PR-401-20200527.0045.gitddf75fd/utils/hwloc/test-hwloc-dump-hwdata/test-hwloc-dump-hwdata.sh.in: file name is too long (max 99); not dumped
```

jsquyres:
Ok, here's the key:

```
${TAR-tar} chof - "$tardir" ...
```

The `o` option is equivalent to `--old-archive`, which means `--format=o7`.

Hence, it's defaulting back down to the ancient tar format (even though you have the new `tar`).

How does this work on Ubuntu 16?  I don't know, and I'm out of time at the moment.  ☹️

bgoglin:
I fails on Ubuntu 16.04 too (tar 1.28 and am 1.15).

bgoglin:
It actually fails on my bleeding-edge Debian too, I am getting the "o" option too.

bgoglin:
And it fails on ubuntu 18.04 too (automake 1.15.1 tar 1.29). I can't test Ubuntu 20 or RHEL 8 right now.

jsquyres:
I must have previously tested incorrectly on Ubuntu 18.04 -- I confirm what you said above: it doesn't work, either.

I don't see a public interface to override the `chof` that Automake puts in as arguments to `tar`.

So perhaps the PR is just a loser; we should close it, because hwloc definitely has filenames that are over 99 characters in length, and the old v7 tar format doesn't allow them.  Of the options that Automake allows, it seems that hwloc needs `ustar` or newer.

PR #404, opened 2020-06-09
Add tests for RSMI plugin

miketxli:
Change-Id: I600e7ea6f4b7e0b71722bca98c605c436316ac51
Signed-off-by: Mike Li <Tianxinmike.Li@amd.com>

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -28,6 +28,7 @@ include_hwloc_HEADERS = \
         hwloc/cuda.h \
         hwloc/cudart.h \
         hwloc/nvml.h \
+        hwloc/rsmi.h \
         hwloc/plugins.h \
         hwloc/gl.h \
         hwloc/rename.h \
--- a/include/hwloc/rename.h
+++ b/include/hwloc/rename.h
@@ -510,6 +510,12 @@ extern "C" {
 #define hwloc_nvml_get_device_osdev HWLOC_NAME(nvml_get_device_osdev)
 #define hwloc_nvml_get_device_osdev_by_index HWLOC_NAME(nvml_get_device_osdev_by_index)
 
+/* rsmi.h */
+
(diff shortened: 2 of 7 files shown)


bgoglin:
Did you test this on a machine with multiple GPUs? We want to make sure we identify GPUs correctly from different ways, but I don't have a working server with more than one AMD GPU.

miketxli:
Yes, I did test it with two GPU's. I have requested to send new cards to you. They have to find out how to it.

miketxli:
After this I will work on to add more new features including XGMI connection. XGMI connects multiple GPU's. Any suggestion for how to add XGMI functionality? I have to go through the list like you mentioned before i.e.:

OpenCLGlobalMemorySize=3087360
OpenCLComputeUnits=68
CUDAGlobalMemorySize=16548352
CUDAL2CacheSize=6144
CUDAMultiProcessors=80
CUDACoresPerMP=64
CUDASharedMemorySizePerMP=48

bgoglin:
Thanks a lot for the shipment. In case it helps, Brad Benton from AMD Research@Austin was my contact who made AMD send some Fiji cards to us in May 2016. We should rather discuss this on my inria.fr email address.

Regarding XGMI, what do you want to expose? That's the interconnect between GPUs, right? If you want to expose the individual links between GPUs, it sound similar to what we want to do for NVLink between NVIDIA GPUs. The current plan (in the nvlink branch) is to expose that through a bandwidth matrix between GPU objects. The big question is which GPU objects. It could be PCI devices, which works very well, but PCI devices isn't what applications actually refer to. It could be OS devices such as rsmiX or nvmlX, but why would those get a distance matrix while cuda and OpenCL OS corresponding OS devices wouldn't? I am tempted to think of doing PCI and having helpers functions to translate those into OS devices.

If you have other information that is purely local to a single GPU, info attribute like those in OpenCL or CUDA are usually the generic way to expose them.

miketxli:
Let me take a look of that branch.

Thanks

Mike

PR #409, opened 2020-07-22
Lstopo tikz

clementFoyer:
I wanted a proper output for inclusion into a LaTeX document and from how straight forward it was to change from a SVG to a tikz picture, I thought it would be useful to have it directly available.

However, I have two issues:
- How can I enable the colors? I tried to define the `declare_color` function, but it seems to never be called.
- Where can I add a flag that insert a backslash in front of the _sharp_ character to have it escaped for tikz?

The result file creates a standalone tex file which can be compiled and depends on `tikz` and `xcolor`. The file can be also be easily included into a latex project.

### DIFF
--- a/utils/lstopo/lstopo-no-graphics.1in
+++ b/utils/lstopo/lstopo-no-graphics.1in
@@ -2,6 +2,7 @@
 .\" Copyright © 2009-2020 Inria.  All rights reserved.
 .\" Copyright © 2009-2010 Université of Bordeaux
 .\" Copyright © 2009-2010 Cisco Systems, Inc.  All rights reserved.
+.\" Copyright © 2020 Hewlett-Packard Enterprise.  All rights reserved.
 .\" See COPYING in top-level directory.
 .TH LSTOPO "1" "%HWLOC_DATE%" "%PACKAGE_VERSION%" "%PACKAGE_NAME%"
 .SH NAME
@@ -445,6 +446,11 @@ Output an ASCII art representation of the map
 (formerly called \fBtxt\fR).
 If outputting to stdout and if colors are supported on the terminal,
 the output will be colorized.
+.
+.TP
+.B tikz or tex
+Output a LaTeX tikzpicture representation of the map that can be
+compiled with a LaTeX compiler.
 .
 .TP
 .B fig
--- a/utils/lstopo/lstopo-tikz.c
+++ b/utils/lstopo/lstopo-tikz.c
@@ -1,0 +1,168 @@
+/*
+ * Copyright © 2020 Hewlett-Packard Enterprise.  All rights reserved.
+ * See COPYING in top-level directory.
+ */
+
+#include "private/autogen/config.h"
+#include "hwloc.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#include "lstopo.h"
+
+#define TIKZ_TEXT_WIDTH(length, fontsize) (((length) * (fontsize))/2.2)
+#define TIKZ_FONTSIZE_SCALE(size) (((size) * 11) / 9)
+
+#define TIKZ_FONTFAMILY_ENV "LSTOPO_TIKZ_FONTFAMILY"
+
+static const char *
+tikz_get_font_family(void)
+{
+  /* Authorized values (case insensitive): sf, rm, tt */
+  char *font_request = getenv(TIKZ_FONTFAMILY_ENV);
+  if (!font_request || !(*font_request))
+    return ""; /* use latex default font family */
+  else if (!strcasecmp(font_request, "sf"))
+    return "\\sffamily";
+  else if (!strcasecmp(font_request, "rm"))
+    return "\\rmfamily";
+  else if (!strcasecmp(font_request, "tt"))
+    return "\\ttfamily";
+  else
+    return NULL;
+}
+
+static int
+tikz_declare_color(struct lstopo_output *loutput, struct lstopo_color *lcolor)
+{
+  int r = lcolor->r, g = lcolor->g, b = lcolor->b;
+  FILE *file = loutput->backend_data;
+
+  /* TODO: add random seed in name to avoid collisions
+   * name format proposed: hwloc-color-<seed>-<r>-<g>-<b>
+   */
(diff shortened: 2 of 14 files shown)


bgoglin:
Hello Clement. Your output_tikz() function should call declare_colors() so that your declare_color() backend function is called before using those colors.
I am not sure we want to hack the core lstopo to insert specific escape characters. XML escapes in its own output functions, can't you do the same by modifying the text buffer before outputting it?

clementFoyer:
Not sure whether the output extension should be `.tex` or `.tikz`, any? Both?

bgoglin:
I don't know, it's really up to you, depending on what convenient way you find to use the output file in latex later.

utils/lstopo/lstopo-no-graphics.1in
```diff
@@ -447,6 +448,11 @@ If outputting to stdout and if colors are supported on the terminal,
 the output will be colorized.
 .
 .TP
+.B tikz or tex
+Output a LaTeX tikzpicture representation of the map that can be
+compiled with a LaTeX compiler to generate a standalone image.
```
bgoglin, line 453: do we really need "to generate a standalone image" since the code may be inserted in the latex document?

utils/lstopo/lstopo-tikz.c
```diff
@@ -0,0 +1,154 @@
...
+    fprintf(stderr, "Failed to open %s for writing (%s)\n", filename, strerror(errno));
+    return -1;
+  }
+
+  font_family = tikz_get_font_family();
+  if (!font_family) {
+    /* TODO: choose what to do on error.
+     * For now, silently ignore it
+     */
+    font_family = "";
```
bgoglin, line 121: I thought you would printf("\\%sfamily", envvar). If getting NULL instead, just fprintf an error message, fclose(output) if not stdout, and return -1.

utils/lstopo/lstopo-tikz.c
```diff
@@ -0,0 +1,154 @@
...
+  loutput->methods = &tikz_draw_methods;
+  loutput->backend_data = output;
+
+  /* recurse once for preparing sizes and positions */
+  loutput->drawing = LSTOPO_DRAWING_PREPARE;
+  output_draw(loutput);
+  loutput->drawing = LSTOPO_DRAWING_DRAW;
+
+  fprintf(output, "\\documentclass{standalone}\n\\usepackage{tikz}\n\\usepackage{xcolor}\n");
+  declare_colors(loutput);
```
bgoglin, line 134: don't you need something like this before and after declare colors?
%%%% If inserting in another document, the following lines below must be copied before \begin{document} %%%%
%%%% End of lines needed before \begin{document} %%%%%

utils/lstopo/lstopo-tikz.c
```diff
@@ -0,0 +1,154 @@
...
+  fprintf(output, "\\begin{document}\n");
+  fprintf(output, "\n%%%%%%%%%% END OF HEADER %%%%%%%%%%\n\n");
```
bgoglin, line 136: %%%% If inserting in another document, this is the actual source code of the picture %%%%

utils/lstopo/lstopo-tikz.c
```diff
@@ -0,0 +1,154 @@
...
+                  "outer sep=0pt,inner sep=0pt,font=\\fontsize{%d}{%d}\\selectfont%s}]\n",
+                  loutput->fontsize, loutput->fontsize + loutput->linespacing, font_family);
+  fprintf(output, "\t\\clip (0,0) rectangle (%u,%u);\n", loutput->width, loutput->height);
+
+  /* ready */
+  lstopo_prepare_custom_styles(loutput);
+
+  output_draw(loutput);
+  fprintf(output,"\\end{tikzpicture}\n");
+  fprintf(output, "\n%%%%%%%%%% BEGIN OF FOOTER %%%%%%%%%%\n\n\\end{document}\n");
```
bgoglin, line 148: %%%% End of actual source code of the picture to insert in another document %%%%%

utils/lstopo/lstopo.c
```diff
@@ -337,7 +338,7 @@ void usage(const char *name, FILE *where)
 #endif
 		  ".\n");
 
-  fprintf (where, "Supported output file formats: console, ascii, fig"
+  fprintf (where, "Supported output file formats: console, ascii, tikz, tex, fig"
```
bgoglin, line 341: Only put tikz here, not the tex alias.

utils/lstopo/lstopo.c
```diff
@@ -512,6 +513,7 @@ enum output_format {
   LSTOPO_OUTPUT_CONSOLE,
   LSTOPO_OUTPUT_SYNTHETIC,
   LSTOPO_OUTPUT_ASCII,
+  LSTOPO_OUTPUT_TIKZPICTURE,
```
bgoglin, line 516: TIKZ is shorted :)

utils/lstopo/lstopo-no-graphics.1in
clementFoyer: Changed for `Output a LaTeX tikzpicture representation of the map that can be compiled with a LaTeX compiler`

utils/lstopo/lstopo-tikz.c
clementFoyer: Done.

utils/lstopo/lstopo-tikz.c
clementFoyer: The idea was to have a label easy to parse so the user and to consider he/she would know what to do if he/she decided to edit it manually/extract the image themselves.

But done.

utils/lstopo/lstopo-tikz.c
clementFoyer: Done.

utils/lstopo/lstopo-tikz.c
clementFoyer: Done.

utils/lstopo/lstopo.c
clementFoyer: Done.

utils/lstopo/lstopo.c
clementFoyer: Done.

PR #415, opened 2020-08-12
Allow to import support bits from XML topology

bgoglin:
With this PR, the "support" bits are always exported to hwloc XML, and they may be imported if new flag HWLOC_TOPOLOGY_FLAG_IMPORT_SUPPORT is set with hwloc_topology_set_flags().


bgoglin:
@rhc54 Would this help for OMPI, even if it comes way too late?

rhc54:
😄 Yes! Still useful for PMIx, so thanks!

bgoglin:
One thing I am wondering: if we pass the flag in hwloc 2.3 import, but the XML was exported by hwloc 2.2 (without support bits), we'll get no support bits (meaning nothing is supported). You'd need to know you have to fallback on your own code.
I need to find a way to report that clearly, either by failing to load, or by adding something else (could be another support bit, but it doesn't really fit in the current arrays).

rhc54:
True, but I thought you were going to add something to the XML string that identified its source version? People could use that anyway as the string has changed a couple of times - might help going forward.

bgoglin:
The XML versioning is only used internally for 1.x/2.x compatibility on import. There's no versioning between 2.[0123] so far, as everybody may import from everybody (they'll just ignore what they don't understand).
Application may read the hwlocVersion string in info attribute but parsing it isn't very convenient (could be "2.1.1rc1-git").
I think I am going to add a new support->misc array with "imported_support" feature flag.

bgoglin:
This will be in 2.3. the way to use it is:
```
#if HWLOC_API_VERSION >= 0x20300
   hwloc_topology_set_flags(topo, <other flags> | HWLOC_TOPOLOGY_FLAG_IMPORT_SUPPORT):
#endif
[...]
   hwloc_topology_load(topo);
[...]
   support = hwloc_topology_get_support(topo);
#if HWLOC_API_VERSION >= 0x20300
   if (support->misc->imported_support)
      <you got the remote support bits>
#endif
```

PR #418, opened 2020-08-24
lstopo/tikz: Update auto-completion with tikz output format

clementFoyer:
The PR from July 23, was lacking the update of the auto-completion of the output format. This commit fixes it. As the man page states both tikz and tex, should `tex` be added as well?


bgoglin:
Thanks. tikz is enough for me, tex is only here for convenience.

PR #422, opened 2020-09-18
Memory attributes

bgoglin:
* API
  + Add hwloc/memattrs.h for exposing latency/bandwidth information between
    initiators (CPU sets for now) and target NUMA nodes.
    - When available, bandwidths and latencies are read from the ACPI HMAT
      table exposed by Linux kernel 5.2+.
    - Attributes may also be customized to expose user-defined performance
      information.
  + Add hwloc_get_local_numanode_objs() for listing NUMA nodes that are
    local to some locality.
* Tools
  + hwloc-info and hwloc-calc have new --local-memory --local-memory-flags
    and --best-memattr options for reporting local memory nodes and filtering
    by memory attributes.
  + hwloc-bind has a new --best-memattr option for filtering by memory attributes
    among the memory binding set.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -27,6 +27,8 @@ Version 2.3.0
       table exposed by Linux kernel 5.2+.
     - Attributes may also be customized to expose user-defined performance
       information.
+  + Add hwloc_get_local_numanode_objs() for listing NUMA nodes that are
+    local to some locality.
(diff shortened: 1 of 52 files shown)


bgoglin:
We usually rebase before merge, but this branch has 21 commits and the summary in the merge commit might be useful to see in the log.

PR #428, opened 2020-10-20
Move PGI compiler check before GNU

jjhursey:
* PGI compiler has a GNU compatability mode so `__GNUC__` is defined by
   `pgcc`. So first check for `__PGI` before checking for GNU.

### DIFF
--- a/config/hwloc_check_vendor.m4
+++ b/config/hwloc_check_vendor.m4
@@ -12,6 +12,7 @@ dnl Copyright © 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright © 2011      Cisco Systems, Inc.  All rights reserved.
 dnl Copyright © 2015-2020 Inria.  All rights reserved.
+dnl Copyright © 2020      IBM Corporation.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


jjhursey:
Configured with `./configure --prefix=$HOME/local/hwloc-pgi CC=pgcc CXX=pgc++ FC=pgfortran` (on `ppc64le`, but that shouldn't matter for this check).

Before the change configure detected the PGI compiler as "gnu"
```
###
### Configuring hwloc core
###
checking hwloc building mode... standalone
configure: hwloc builddir: /smpi_dev/mpiczar/local/src/hwloc-2.2.0
configure: hwloc srcdir: /smpi_dev/mpiczar/local/src/hwloc-2.2.0
checking for hwloc version... 2.2.0
checking if want hwloc maintainer support... disabled
checking for hwloc directory prefix... (none)
checking for hwloc symbol prefix... hwloc_
checking for pgcc option to accept ISO C99... (cached) none needed
checking size of void *... 8
checking which OS support to include... Linux
checking which CPU support to include... unknown
checking size of unsigned long... 8
checking size of unsigned int... 4
checking for the C compiler vendor... gnu
```

After this commit, it correctly detects it as "portland group"
```
###
### Configuring hwloc core
###
checking hwloc building mode... standalone
configure: hwloc builddir: /tmp/jjhursey/hwloc
configure: hwloc srcdir: /tmp/jjhursey/hwloc
checking for hwloc version... 2.4.0a1-git
checking if want hwloc maintainer support... disabled
checking for hwloc directory prefix... (none)
checking for hwloc symbol prefix... hwloc_
checking for pgcc option to accept ISO C99... (cached) none needed
checking size of void *... 8
checking which OS support to include... Linux
checking which CPU support to include... unknown
checking size of unsigned long... 8
checking size of unsigned int... 4
checking for the C compiler vendor... portland group
```

bgoglin:
Thanks Josh. Is this a new feature in pgcc? Or something most people never enable?

bgoglin (approved):

jjhursey:
I think it's been there for a few releases at least. In Open MPI we made the change a while back to the [similar m4 file](https://github.com/open-mpi/ompi/blob/master/config/opal_check_vendor.m4#L113-L116). I just noticed it today because of an odd configuration/setup I had with Open MPI's new 3rd party configure logic. It was trying to pass down GNU only warning CLI options into hwloc, which PGI doesn't handle.

It might be more apparent when using the new v20.x `pgcc` or the older v19.x `pgcc18`, but I would need to dig into that a bit further to confirm.

PR #432, opened 2020-11-05
API for CPU kinds aka heterogeneous/hybrid CPUs

bgoglin:
Most of this code was merged in https://github.com/open-mpi/hwloc/pull/432/commits/7301168792a4dbd34c568e2f20d1ced38194fbe3
Remaining stuff include support for upcoming Linux sysfs cpu types files. Those patches are merged in upstream Linux yet but it looks like we'll be able to include that support anyway, I am just waiting in case the sysfs layout is modified in newer patches.


bgoglin:
Entirely merged in master in commit 8ef4029d1aac7b8c8d803d13b6f276624fc4cac6

PR #435, opened 2020-11-26
XGMIBandwidth matrix

bgoglin:
@miketxli Here's a totally untested draft of XGMI distance matrix. I just put 100 in the matrix when your code found a XGMI peer (you said InfinityFabric links are 100GB/s, right). The code isn't ready to merge (I'd need to export the distances_add() function to plugins, right now it will only work when the RSMI backend is not a separate plugin), but I'd like you to test/fix/discuss the code in the meantime. I am working on NVLink and oneAPI support at the same time.

If you run lstopo -v, you should get a distance matrix between OSDev at the bottom of the output. The numbers on the first row and first column are the OS Devices logical indexes (those you see as L#1 in lstopo --only osdev -v | grep rsmi).

The diagonal would keep 0 which is a normal because because there's no XGMI link to myself, but also a bit strange since the bandwidth to myself is likely higher than the XGMI link bandwidth. I don't see what else to put in there. UINT64_MAX would case horribly large ugly values to be displayed in lstopo.

Instead of 100 or 100000 for bandwidth in GB/s or MB/s, we could also just put 1 in the matrix when there's a link. If XGMI ever support multiple links between a pair of GPUs (looks like NVLink already does), we'd still be able to put 2 instead of 1. The question is whether we could have something more complicated with different kinds of links at some point. That's why I used 100 for a 100GB/s single link here, seemed more flexible than 0 or 1.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,8 +38,10 @@ Version 2.5.0
     Thanks to Anara Kozhokanova, Tim Cramer and Erich Focht for the help.
   + Add a NVLinkBandwidth distances structure between NVIDIA GPUs
     (and POWER processor or NVSwitches) in the NVML backend.
-    See Topology Attributes: Distances, Memory Attributes and CPU Kinds
-    in the documentation.
+  + Add a XGMIBandwidth distances structure between AMD GPUs
+    in the RSMI backends.
+  + See Topology Attributes: Distances, Memory Attributes and CPU Kinds
+    in the documentation for details about these new distances.
 * Build
   + Add --with-cuda-version=<version> or look at the CUDA_VERSION
     environment variable to find the appropriate CUDA pkg-config files.
--- a/doc/hwloc.doxy
+++ b/doc/hwloc.doxy
@@ -2091,6 +2091,12 @@ Distances structures currently created by hwloc are:
 <dt>NUMALatency (Linux, Solaris, FreeBSD)</dt>
 <dd>This is the matrix of theoretical latencies described above.
 </dd>
+<dt>XGMIBandwidth (RSMI)</dt>
+<dd>This is matrix of unidirectional XGMI bandwidths between AMD GPUs (in MB/s).
+It contains 0 when there is no direct XGMI link between objects.
+Values on the diagonal are artificially set to very high so
+that local access always appears faster than remote access.
+</dt>
 <dt>NVLinkBandwidth (NVML)</dt>
 <dd>This is the matrix of unidirectional NVLink bandwidths between
 NVIDIA GPUs (in MB/s).
--- a/hwloc/topology-rsmi.c
+++ b/hwloc/topology-rsmi.c
@@ -15,6 +15,38 @@
 
 #include <rocm_smi/rocm_smi.h>
 
+static int
+hwloc__rsmi_add_xgmi_bandwidth(hwloc_topology_t topology,
+                               unsigned nbobjs, hwloc_obj_t *objs, hwloc_uint64_t *bws)
+{
+  void *handle;
+  int err;
+
+  handle = hwloc_backend_distances_add_create(topology, "XGMIBandwidth",
+                                              HWLOC_DISTANCES_KIND_FROM_OS|HWLOC_DISTANCES_KIND_MEANS_BANDWIDTH,
+                                              0);
+  if (!handle)
+    goto out;
+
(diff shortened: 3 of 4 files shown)


bgoglin:
Here's what I see for NVLink:
```
$ utils/lstopo/lstopo-no-graphics --only osdev -v | grep nvml
GPU L#5 (Backend=NVML GPUVendor="NVIDIA Corporation" GPUModel="Tesla P100-SXM2-16GB" NVIDIASerial=0322816175083 NVIDIAUUID=GPU-81504486-a0dc-e7c9-4a2e-2a1077df8a42) "nvml0"
GPU L#7 (Backend=NVML GPUVendor="NVIDIA Corporation" GPUModel="Tesla P100-SXM2-16GB" NVIDIASerial=0322716069829 NVIDIAUUID=GPU-7c2fa792-8540-b761-a5d9-41769fe906d8) "nvml1"
GPU L#17 (Backend=NVML GPUVendor="NVIDIA Corporation" GPUModel="Tesla P100-SXM2-16GB" NVIDIASerial=0322816175517 NVIDIAUUID=GPU-f633f2f8-1587-fa7e-a77f-fd3b7498e16b) "nvml2"
GPU L#19 (Backend=NVML GPUVendor="NVIDIA Corporation" GPUModel="Tesla P100-SXM2-16GB" NVIDIASerial=0322816174491 NVIDIAUUID=GPU-953bb196-b5f9-81cf-1601-c0d90c795ea0) "nvml3"
```
Hence GPUs are logical indexes 5/7/17/19 for hwloc.
```
$ utils/lstopo/lstopo-no-graphics -v
[...]
Relative bandwidth matrix (name NVLinkBandwidth kind 9) between 4 OSDevs (depth -6) by logical indexes:
  index     5     7    17    19
      5     0  40000     0     0
      7  40000     0     0     0
     17     0     0     0  40000
     19     0     0  40000     0
```

miketxli:
Thanks for update. Harish, my manager will setup a meeting with you to discuss, probably next week after US thanksgiving.

bgoglin:
@miketxli The core changes for adding GPU distances are pretty much ready. I updated the nvml and rsmi code to insert insert-GPU bandwidth matrices. Can you test the tarball at https://ci.inria.fr/hwloc/job/basic/job/PR-435/ and report what "lstopo --distances" shows?

miketxli:
@bgoglin Hi Brice, Thanks for the contributions. I'm working on some urgent issues right now and will get to this after that.

miketxli:
@bgoglin Hi Brice, sorry about delay. I finally get chance to do this task. I'm setting up a testing system and will test this.

miketxli:
@bgoglin Hi Brice. Sorry about delay. This is the busy time. I'm working on multiple projects at same time. I final get a working system with two GPUs connected with xGMI. I have done test, but got the following results:
Relative latency matrix (name NUMALatency kind 5) between 2 NUMANodes (depth -3) by logical indexes:
 index     0     1
     0    10    21
     1    21    10
The number and description don't seems correct.

bgoglin:
Hello @miketxli. It looks like no XGMIBandwidth matrix was created. Can you post "lstopo -v | grep -i rsmi" to check whether bandwidth were added in the rsmi object "XGMIPeers" attribute?

miketxli:
It seems  that a lstopo-no-graphics without rsmi was called, even so I have called from utils/lstopo with ./lstopo-no-graphics -v. I have to do sudo make install to fix it. It works now.
Relative bandwidth matrix (name XGMIBandwidth kind 9) between 2 OSDevs (depth -6) by logical indexes:
   index   rsmi0   rsmi1
   rsmi0 1000000     100
   rsmi1     100 1000000

miketxli:
I think I like a bit more to use 0 instead 1000000 and 1 for 100 as you have suggested. Then we can put number of links if there are multiple links between 2 GPUs.

bgoglin:
Thanks for testing, it looks like the core works, at least.

I think I am going to keep GB/s (and make it 200 if there are 2 links) because it's more flexible if we ever create larger matrices involving different interconnects (PCI, NUMA, etc). But I'll provide a helper to convert this raw matrix into the number of links since it's quite easy to do : put 0 on the diagonal, divide other values by the greatest common divisor (if all links are doubled, we'll report single links but that's OK).

By the way,  my 1000000 on the diagonal is wrong, it should have been 1000 (I am going to use GB/s for this bandwidth).

miketxli:
I'm good with GB/s too. Thanks

PR #436, opened 2020-12-02
NVLink bandwidth matrix

bgoglin:
TODO official plugin API for adding matrices
TODO is it ok to consider non-PCI bridges as PCI devices?
TODO helper to translates matrice from nvml objects to cuda objects, with flag to remove non-GPU objects?
TODO switch some current printf to hwloc_debug

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -36,6 +36,10 @@ Version 2.5.0
   + Expose NEC Vector Engine cards on Linux as OS devices of subtype
     "VectorEngine" and name "ve0", etc.
     Thanks to Anara Kozhokanova, Tim Cramer and Erich Focht for the help.
+  + Add a NVLinkBandwidth distances structure between NVIDIA GPUs
+    (and POWER processor or NVSwitches) in the NVML backend.
+    See Topology Attributes: Distances, Memory Attributes and CPU Kinds
+    in the documentation.
 * Build
   + Add --with-cuda-version=<version> or look at the CUDA_VERSION
     environment variable to find the appropriate CUDA pkg-config files.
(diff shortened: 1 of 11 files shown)


sthibaul:
Looking at the support for NVLink, I'm thinking that the information I'd like to have is whether two GPUs are connected directly, or with others (and which ones) through an NVSwitch.

sthibaul:
I see that the bandwidth matrix provides the information: when a direct link is available, a bandwidth is there, but for indirect links, AIUI there is another entry in the matrix for the NVSwitch, but also for GPU-to-CPU PowerNVL bridge, we'd need to have something to distinguish the two.

Probably people will like to also have a transitive-closure matrix, to directly get the available bandwidth between two GPUs (even if it goes through a switch shared with other GPUs)

bgoglin:
The discussion is split between this issue and the other one about XGMI but here are the ideas gathered so far. After you get the distances structure with get(), you'll have some functions to transform it:
* switch from bandwidth to 1 for direct link and 0 for no link
* merge NVswitch objects into a single object. By default we see 2 sets of subswitches without connections between them. I wonder if I should merge them by default (the PCI locality of the switch would be wrong, but it doesn't really matter).
* remove NVswitch and/or CPU ports (generic API to filter distances with a bitmask of object indexes)
* replace objects: nvml/rsmi -> cuda/opencl/PCI. Could also be a generic API to replace any object given by type/name with another type/name with same locality.

Transitive closure requires a careful definition because GPUs cannot route NVLink traffic (you need an intermediate kernel for forwarding). We may want a strict mode where only things that can route traffic are considered (NVSwitches), and a more relaxed mode where user may have to help.

sthibaul:
> Transitive closure requires a careful definition because GPUs cannot route NVLink traffic (you need an intermediate kernel for forwarding). We may want a strict mode where only things that can route traffic are considered (NVSwitches),

Yes, I meant closure through whatever can actually route (thus a NVSwitch for instance, possibly something else in the future)

PR #439, opened 2020-12-30
Fix GCC checks to use the vendor name

jjhursey:
* The PGI compiler is identified as a GCC compiler (so `$GCC == yes`)
   This can cause problems if the user accidentally adds `--enable-picky`
   or if the `$GCC` variable is used to check if it is safe to add GCC
   specific compiler flags.
 * Instead of using `$GCC` we can use the vendor string returned by
   `_HWLOC_C_COMPILER_VENDOR`.

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -13,6 +13,7 @@ dnl Copyright © 2006-2017 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright © 2012  Blue Brain Project, BBP/EPFL. All rights reserved.
 dnl Copyright © 2012       Oracle and/or its affiliates.  All rights reserved.
 dnl Copyright © 2012  Los Alamos National Security, LLC. All rights reserved.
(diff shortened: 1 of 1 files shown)


jjhursey:
I stumbled on this one by accident today while bouncing between GNU and PGI compilers. It shouldn't impact too many folks, but good cleanup to do.

bgoglin:
Hello Josh, can you remove the existing call to _HWLOC_C_COMPILER_VENDOR([hwloc_c_vendor]) a bit later in the same HWLOC_SETUP_CORE?

jjhursey:
Done. I just repushed.

bgoglin (approved):

bgoglin:
Thanks Josh, I backported to v2.4 too. If you want it in OMPI, you can take the whole v2.4 branch (there 4 bugfixes since 2.4.0, other commits don't touch the library code). If you need it in another stable branch, let me know.

PR #448, opened 2021-02-03
Fix xl compiler check

awlauria:
- By default newer xlc compilers only define __ibmxl__ now.

https://www.ibm.com/support/knowledgecenter/en/SSXVZZ_13.1.6/com.ibm.xlcpp1316.lelinux.doc/compiler_ref/xlmacros.html

- Most compilers define the __GNU__ macro, so put it at the bottom
as a catch-all.

Tested with gcc, xlc, pgi and clang.


bgoglin:
Thanks, I'll merge tomorrow and backport to 2.4. I don't think it even deserves a 2.4.1rc3 before actual release.

Issue #366, opened 2019-08-19
x86 changes the process affinity on Windows during load()


ivankochin:
### System information

| Title                 | Value                                        |
|-----------------|----------------------------------|
| hwloc version  | 2.0.4                                         |
| OS                   | Windows 10                             |
| Hardware        | 2 x Intel(R) Xeon(R) Gold 6152 |

### Details of the problem

Function `hwloc_topology_load()` changes process affinity masks on machines with multiple processors groups. 

Test case:
``` c++

#include <hwloc.h>
#include <iostream>
#include <windows.h>

int main(void)
{
    DWORD_PTR pam, sam;
    GetProcessAffinityMask( GetCurrentProcess(), &pam, &sam );
    std::cout << "Process affinity mask before: " << std::hex << pam << std::endl;

    hwloc_topology_t topology;
    hwloc_topology_init(&topology);
    hwloc_topology_load(topology);

    GetProcessAffinityMask( GetCurrentProcess(), &pam, &sam );
    std::cout << "Process affinity mask after: " << std::hex << pam << std::endl;

    return 0;
}

```
When I run it with an affinity mask equal to 1, I get the following output:
```
Process affinity mask before: 1
Process affinity mask after: fffffffffff
```
As I can see, process affinity mask changes to full processors group affinity mask. This behavior appears only on machines that have more than 1 processors groups.

bgoglin:
Hello

Can you set HWLOC_COMPONENTS=-x86 in the environment to check whether this is caused by the x86 backend (it binds on each core to perform discovery on each of them)? This backend isn' very useful, it just adds a bit more information to the topology, disabling it is fine.

One thing that isn't clear on Windows is the relation between process and thread binding. On Linux, there's no "process-wide" binding, it's just the combination of the bindings of individual threads in that process. On Windows, there are API for both. if I am reading the code correctly, when there are multiple proc groups, we only have thread binding support. Can you check the thread binding in your code too? Using both GetThreadGroupAffinity()?

By the way, how do you switch processor groups on/off? Do they get auto-enabled because you have 2x 22 dual-threaded cores > 64 PUs?

sthibaul:
IIRC, on windows a process can be only in one proc group. That could mean that even just calling SetThreadGroupAffinityMask(), which was assumed to have effect only on the current thread, will change the whole process' group binding, since the process will then have to go to the group requested for the current thread. We thus might have to introduce a more complex hook to properly restore not only the thread binding but also the process binding

[...]

PR #452, opened 2021-02-18
hwloc/windows.h for querying Windows processor groups

bgoglin:
Windows processor groups are not a physical hierarchical level (although it may be exposed by hwloc if it's different from NUMA nodes, packages, etc) but it has an impact of application because binding must remain inside a single group. Expose those groups explicitly so that applications know what their binding masks should be restricted to.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,8 @@ Version 2.5.0
 -------------
 * API
   + Add hwloc/windows.h to query Windows processor groups.
+* Tools
+  + lstopo now has a --windows-processor-groups option on Windows.
 * Build
   + Add --with-cuda=<dir> to specify the CUDA installation path
     (and its NVML and OpenCL components) without environment variables.
--- a/NEWS
+++ b/NEWS
@@ -19,6 +19,8 @@ bug fixes (and other actions) for each version of hwloc since version
 
 Version 2.5.0
 -------------
+* API
+  + Add hwloc/windows.h to query Windows processor groups.
 * Build
   + Add --with-cuda=<dir> to specify the CUDA installation path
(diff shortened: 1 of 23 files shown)


bgoglin:
@ivankochin Tarballs from this PR is available at https://ci.inria.fr/hwloc/job/basic/job/PR-452/
lstopo doesn't display processor groups by default anymore, you have to pass --windows-processor-groups (or pass -v if there are multiple processor groups in the system).
The API now enforces that the current topology matches the system topology (i.e. not loaded from another machine through XML). This should be fine since the goal of this API is to help binding anyway.

ivankochin:
I can't find HWLOC packages for Windows by the link shared above. 
Could you please provide such packages or share the build instructions to compile the HWLOC on Windows using the right environment?

bgoglin:
Sorry, I forgot that the PR builder doesn't build windows binaries. I made those available in https://ci.inria.fr/hwloc/job/extended-customPR/14/

ivankochin:
I tested this patch on several machines and it seems good. Thanks for providing this custom build.

Could you please estimate when this feature is planned to be released? Do you have plans to release it as a part of HWLOC 2.5?
Also, I have the same questions regards #366.

bgoglin:
I just merged this in master for 2.5. The plan for 2.5 isn't totally clear yet, but it should be released in April. The plan is to have some NVIDIA/AMD GPU interconnect matrices (pretty much ready) and the new topology flag to avoid binding issues (#366). There are corner cases in cpukinds that are not ready for 2.5 yet, and they are mitigated by the binding fixes anyway, hence I don't plan to delay 2.5 because of them.

PR #461, opened 2021-04-05
Make -ltermcap the last option for ncurses

haampie:
On macOS there seems to be a symlink from /usr/lib/libtermcap.dylib to
libncurses.5.4.dylib, which is the first option hwloc currently
considers for detecting ncurses. If you install ncurses with brew or
spack, you don't get this symlink, so hwloc will always link to the
system ncurses instead of the brew/spack version.

By considering libncurses first this should be fixed.

### DIFF
--- a/config/hwloc_internal.m4
+++ b/config/hwloc_internal.m4
@@ -339,7 +339,7 @@ EOF
     chosen_curses=""
     for curses in ncurses curses
     do
-      for lib in "" -ltermcap -l${curses}w -l$curses -ltinfo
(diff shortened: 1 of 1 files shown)


bgoglin:
@sthibaul Does this look fine to you?

sthibaul:
That should be fine, yes.

bgoglin:
@haampie Thanks, this will be in 2.5.0 (not sure there will be a 2.4.2 before).

haampie:
Thanks!

PR #479, opened 2021-08-20
linux: use MPOL_PREFERRED_MANY when available

### LINKED ISSUE #236: MEMBIND_BIND clarification on Linux
BIND without STRICT on Linux uses MPOL_PREFERRED. Only the first bit of the nodeset is used as a preferred node.

If full, next allocations will come from "nearby" nodes instead of the other bits of the nodeset. Not sure how that "nearby" is actually implemented.

In contrast, BIND with STRICT does respect the nodeset, but it never falls back outside of the nodeset.

bgoglin:
MPOL_PREFERRED is less strict than MPOL_BIND since it falls back
to other node if the given one is full. But it works only with
a single node. Other (non-first physical index) are just ignored.

MPOL_PREFERRED_MANY in 5.15 fixes this.

Use MPOL_PREFERRED_MANY by default for membind. If it fails,
try MPOL_PREFERRED. If it works, assume MPOL_PREFERRED_MANY isn't
supported in the current kernel.

Closes #236

Signed-off-by: Brice Goglin <Brice.Goglin@inria.fr>

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,6 +24,9 @@ Version 2.6.0
     high-performance cores (firestorm) on Apple M1 on Mac OS X.
   + Use sysfs CPU "capacity" to rank hybrid cores by efficiency
     on Linux when available (mostly on recent ARM platforms for now).
+  + Improve HWLOC_MEMBIND_BIND (without the STRICT flag) on Linux kernel
+    >= 5.15: If more than one node is given, the kernel may now use all
+    of them instead of only the first one before falling back to others.
 * Build
   + Allow to specify the ROCm installation for building the RSMI backend:
     - Use a custom installation path if specified with --with-rocm=<dir>.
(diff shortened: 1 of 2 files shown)


bgoglin:
MPOL_PREFERRED was merged in Linux 5.15-rc1 (commit b27abaccf8e8b012f126da0c2a1ab32723ec8b9f). I'll merge this PR once I get a chance to test that kernel.

bgoglin:
hwloc before this PR or running on kernel < 5.15 on machine with 4+ NUMA node:
```
$ hwloc-bind --membind node:2-3 -- hwloc-bind --get --membind --nodeset
0x00000040 (bind)
```
(we request binding on 2 nodes, the kernel only remembers the first one)

hwloc after this PR on 5.15-rc1:
```
$ hwloc-bind --membind node:2-3 -- hwloc-bind --get --membind --nodeset
0x000000c0 (bind)
```
(both nodes are used for real)

PR #480, opened 2021-08-20
Windows Die and Module (and NumaNodeEx) relationships

bgoglin:
Some future Windows release (TBD) will expose "Die" and "Module" information as well as NUMA nodes spanning mulitple processor groups. Bits are appearing in the onlie doc, we don't know all the details yet, and we don't have a way to test yet.


bgoglin:
Requested some clarification from Microsoft at https://github.com/MicrosoftDocs/feedback/issues/3917

PR #514, opened 2022-01-28
Set ZES_ENABLE_SYSMAN via setenv instead of putenv

jjhursey:
* Setting `ZES_ENABLE_SYSMAN` via `putenv` placed a constant string
   in the environ array which cannot be touched. If the user is
   manipulating that environ array then touching this envar will
   result in a segv.
   - Instead of using `putenv` use `setenv` which will put a copy
     of the constant string in the `environ` array allowing the
     end user to manipulate that array as needed.
   - Note that I could not find a `setenv` function for windows
     so I left a comment and did not touch that code.

### DIFF
--- a/hwloc/topology.c
+++ b/hwloc/topology.c
@@ -3,6 +3,7 @@
  * Copyright © 2009-2022 Inria.  All rights reserved.
  * Copyright © 2009-2012, 2020 Université Bordeaux
  * Copyright © 2009-2011 Cisco Systems, Inc.  All rights reserved.
+ * Copyright © 2022 IBM Corporation.  All rights reserved.
  * See COPYING in top-level directory.
  */
 
@@ -74,13 +75,14 @@ static void hwloc_constructor(void) __attribute__((constructor));
 static void hwloc_constructor(void)
 {
   if (!getenv("ZES_ENABLE_SYSMAN"))
-    putenv((char *) "ZES_ENABLE_SYSMAN=1");
+    setenv("ZES_ENABLE_SYSMAN", "1", 1);
 }
 #endif
 #ifdef HWLOC_WIN_SYS
 BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpReserved)
 {
   if (fdwReason == DLL_PROCESS_ATTACH) {
+    // Windows does not have a setenv, so use putenv
     if (!getenv("ZES_ENABLE_SYSMAN"))
       putenv((char *) "ZES_ENABLE_SYSMAN=1");
   }


jjhursey:
We found this when moving from hwloc 2.4.0 to hwloc 2.7.0 and building OpenPMIx. One of the OpenPMIx tests was crashing [here](https://github.com/openpmix/openpmix/blob/0c5fe71a47a01c43e409c1e5ecbd12e06d3dae7c/src/mca/pfexec/base/pfexec_base_default_fns.c#L204) but only when processing the `ZES_ENABLE_SYSMAN` envar.
 * `strdup`'ing the `ZES_ENABLE_SYSMAN` envar allowed it to pass, but the core problem was how it was being set.
 * Changing hwloc from using `putenv` (which will put the string constant in the `environ`) to using `setenv` (which puts a copy of the string in the `environ`) fixed this issue without any changes to the OpenPMIx codebase.

It looks like [this](https://github.com/open-mpi/hwloc/commit/3108bf38640a9fea1f06e7c81106d166da4d5d66) commit introduced the `ZES_ENABLE_SYSMAN` in the hwloc constructor.

bgoglin:
Thanks Josh, I'll verify this still works fine with oneAPI L0. Hopefully Intel will fix their API at some point and we'll remove these ugly hacks in constructors.

rhc54:
@bgoglin You might want to scrub the HWLOC code and replace all instances of "putenv" with "setenv". Per the man page:

     The putenv() function takes an argument of the form ``name=value'' and is equivalent to:

           setenv(name, value, 1);

     The string pointed to by string becomes part of the environment.  A program should not
alter or free the string, and should not use stack or other transient
     string variables as arguments to putenv().  **The setenv() function is strongly preferred to putenv().**

Emphasis is added by me. Other references make stronger statements about that preference as the environ is not protected as "const" and is frequently altered by programs.

rhc54:
@bgoglin Just as an FYI - it looks like I'll need to blacklist HWLOC v2.5.0 and above for PMIx and PRRTE as we have two problems:

* the one cited here. I could put some code into PMIx/PRRTE to strdup the envars prior to working with them as a workaround, but...
* the CUDA library dependency problem. We have tested and found that unless `--enable-plugins` was used to configure HWLOC, both PMIx and PRRTE will fail to initialize. Neither has a CUDA dependency, but apparently we just can't load HWLOC that way. If we reconfigure HWLOC appropriately, then we are okay - but since we build against external installations of HWLOC that we cannot control, that isn't going to work.

Not sure of the right path forward. For now, we'll just do the blacklist and hopefully come up with a better solution in the future.

bgoglin:
> @bgoglin You might want to scrub the HWLOC code and replace all instances of "putenv" with "setenv".

Don't worry, there are no putenv in libhwloc except this new one for L0. Others are in tools and tests, where putenv() is easier because it works on Windows.

Regarding the CUDA issue, it should only be related to the hwloc configury better detecting CUDA. If problems exists now, they likely existed in the past too, but they weren't visible because CUDA was less often detected. There are no change in the hwloc CUDA backend between 2.4 and 2.5 (except some debug printf).

rhc54:
> Regarding the CUDA issue, it should only be related to the hwloc configury better detecting CUDA. If problems exists now, they likely existed in the past too, but they weren't visible because CUDA was less often detected. There are no change in the hwloc CUDA backend between 2.4 and 2.5 (except some debug printf).

Gotcha - unfortunately, it's a complete blocker for us. I've added the "blacklist" configure code until something can be worked out. Until then, people wanting to use PMIx/PRRTE ((and I guess OMPI v5 by inference) will have to limit themselves to v2.4.x.

bgoglin:
I added some minor bits to your commit and pushed it to master and v2.7, thanks.
I can release 2.7.1 next week, or we can wait for some additional fixes for CUDA, but I have no idea where to start since I don't understand the problem and cannot reproduce it.

rhc54:
> or we can wait for some additional fixes for CUDA, but I have no idea where to start since I don't understand the problem and cannot reproduce it.

I wish I could provide some advice there, but I don't have access to a CUDA-enabled machine and therefore cannot reproduce it myself. Basically, it manifests as a failure when trying to load PMIx with an error about being unable to find libnvidia (I believe that was the library). By enabling plugins, all we really do is allow that plugin to fail to load without the failure impacting core hwloc - and so PMIx can still load.

My suspicion is that you don't quite have your Makefile libs correct - you aren't correctly linking libnvidia into hwloc. Thus, you got better at detecting the CUDA libraries - but unfortunately, they aren't being correctly included in hwloc. It's just a guess, but it probably is something like that. You might talk to Brian or Jeff as they have more experience in it (Brian might have access to CUDA-enabled machines).

jjhursey:
@bgoglin Thanks for fixing up that patch and committing! I see the following two commits to master with it (references for those playing at home:
 * https://github.com/open-mpi/hwloc/commit/e7214879cdc7998ff57a99eb902378cc5684fff6
 * https://github.com/open-mpi/hwloc/commit/ed5da56b1d2091264fe4629ac51109e61264d873

I think a 2.7.1 with this fix would be good, but maybe wait to see if we can get traction on the CUDA configure stuff. I have an idea I want to play with that might help with the CUDA detection issue that was mentioned. I have a container build box that can reproduce fairly easily. I see you filed https://github.com/open-mpi/hwloc/issues/515 to help track that. I'll follow up there.

flx42:
I found this bug while investigating a Slurm crash on Ubuntu 22.04 related to the PMIx plugin: https://bugs.schedmd.com/show_bug.cgi?id=14276, the crash was 100% reproducible since `putenv` in a library that can unloaded through `dlclose` is unsafe as mentioned in #533

Using `setenv` instead will remove the guaranteed crash, but I don't believe it's safe either if libhwloc is loaded in a multi-threaded program. Concurrent calls to `getenv` and `setenv` can crash a program:
https://rachelbythebay.com/w/2017/01/30/env/
https://stackoverflow.com/questions/30476600/how-can-getenv-be-thread-safe

While the race condition is possible, it's probably (very) rare to hit a crash (compared to the `putenv` approach). I think the safest option is for us to set `ZES_ENABLE_SYSMAN=0` in the environment to make the situation a little safer as hwloc won't call `setenv` if the variable is already set.

bgoglin:
Note that setenv() is called in the library constructor. In the vast majority of cases, there's a single thread at that point.
The proper solution is to wait for level-zero to stop relying on an environment variable for enabling sysman. This is under discussion but will take months before being available :/

Issue #515, opened 2022-01-30
enable CUDA/ROCm when not in LD_LIBRARY_PATH?


bgoglin:
Since hwloc 2.5, we autodetect CUDA in /usr/local/cuda. Since 2.6, we also autodetect ROCm under /opt/rocm. And we support custom path with --with-cuda=... --with-rocm=... It works fine but causes our library to depend on NVIDIA/AMD libraries that are not always in LD_LIBRARY_PATH or ld.so.conf.

The CUDA installer seems to install /etc/ld.so.conf.d/cuda-x.conf at least on CentOS but some platforms don't have it, causing end-users to get unexpected mpirun failures because it couldn't find some NVIDIA lib, e.g. ompi#9907 on Amazon EC2

It's quite common to support --with-foo=/my/path/that's/not/in/LD_LIBRARY_PATH so I don't think we're doing anything wrong, but maybe there's a way to warn users that we're enabling something that will require LD_LIBRARY_PATH? May require AC_RUN_IFELSE().

Or maybe people use rpath to hardwire the path of those dependencies?

jjhursey:
We have a build/run container environment that we maintain for Open MPI/OpenPMIx/PRRTE CI. It has CUDA installed, but the library `libnvidia-ml.so.1` is missing (we have the stub in `/usr/local/cuda-11.0/targets/ppc64le-linux/lib/stubs/libnvidia-ml.so`). On the host system the true library is located in `/usr/lib64/`, but in the container it does not. I suspect because we are not using the nvidia container runtime on this machine, but that's somewhat beside the point.

The hwloc `config.log` reports:
```
configure:26463: checking for nvmlInit in -lnvidia-ml
configure:26488: gcc -o conftest -g -O2  -I/usr/local/cuda/include/  -L/usr/local/cuda/lib64/ -L/usr/local/cuda/lib64/stubs/ conftest.c -lnvidia-ml   >&5
configure:26488: $? = 0
configure:26497: result: yes
```

But when I run it shows this error message:
```
shell$ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64/stubs/                         
shell$ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64/      
shell$ lstopo
lstopo: error while loading shared libraries: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
```

[...]

PR #516, opened 2022-02-01
Strengthen the NVML check by adding a runtime check

jjhursey:
* The Nvidia ML library is provided by a stub which can fool
   the linker into thinking that the actual library is on the system
   even when it is not. This can cause a runtime failure on such
   systems.
   - Strengthen the `nvidia-ml` library check by adding a runtime
     check that the resulting program will run not just compile
     and link.

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -13,7 +13,7 @@ dnl Copyright © 2006-2017 Cisco Systems, Inc.  All rights reserved.
 dnl Copyright © 2012  Blue Brain Project, BBP/EPFL. All rights reserved.
 dnl Copyright © 2012       Oracle and/or its affiliates.  All rights reserved.
 dnl Copyright © 2012  Los Alamos National Security, LLC. All rights reserved.
-dnl Copyright © 2020 IBM Corporation.  All rights reserved.
+dnl Copyright © 2020-2022 IBM Corporation.  All rights reserved.
 dnl See COPYING in top-level directory.
 
 # Main hwloc m4 macro, to be invoked by the user
@@ -1163,6 +1163,7 @@ return 0;
 
     # NVML support
     hwloc_nvml_happy=no
+    hwloc_nvml_warning=no
     if test "x$enable_io" != xno && test "x$enable_nvml" != "xno"; then
       echo
       echo "**** NVML configuration"
@@ -1174,9 +1175,30 @@ return 0;
       CPPFLAGS="$CPPFLAGS $HWLOC_NVML_CPPFLAGS"
       tmp_save_LDFLAGS="$LDFLAGS"
       LDFLAGS="$LDFLAGS $HWLOC_NVML_LDFLAGS"
+      tmp_save_LIBS="$LIBS"
       AC_CHECK_HEADERS([nvml.h], [
-        AC_CHECK_LIB([nvidia-ml], [nvmlInit], [HWLOC_NVML_LIBS="-lnvidia-ml"], [hwloc_nvml_happy=no])
+        AC_CHECK_LIB([nvidia-ml],
+                     [nvmlInit],
+                     [AC_MSG_CHECKING([whether a program linked with -lnvidia-ml can run])
+                      HWLOC_NVML_LIBS="-lnvidia-ml"
+                      LIBS="$LIBS $HWLOC_NVML_LIBS"
+                      AC_RUN_IFELSE([
+                        AC_LANG_PROGRAM([[
+#include <stdio.h>
+char nvmlInit ();
+]], [[
+ return nvmlInit ();
+]]
+                       )],
+                       [AC_MSG_RESULT([yes])
+                        hwloc_nvml_warning=no],
+                       [AC_MSG_RESULT([no])
+                        hwloc_nvml_warning=yes],
+                       [AC_MSG_RESULT([don't know (cross-compiling)])
+                        hwloc_nvml_happy=maybe])],
+                     [hwloc_nvml_happy=no])
       ], [hwloc_nvml_happy=no])
+      LIBS="$tmp_save_LIBS"
       CPPFLAGS="$tmp_save_CPPFLAGS"
       LDFLAGS="$tmp_save_LDFLAGS"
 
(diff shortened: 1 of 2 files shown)


jjhursey:
Related to Issue https://github.com/open-mpi/hwloc/issues/515

This is a specific commit for the `libnvidia-ml.so` library that I described in the ticket [here](https://github.com/open-mpi/hwloc/issues/515#issuecomment-1026001916).

I'm putting this in draft for the moment because I want to see if @bgoglin approves of the direction, and if there is a desire to add a `--skip-nvml-runtime-check` configure option that will skip the `AC_RUN_IFELSE`.

Note that this resolves the specific issue that I was able to reproduce. There might be more places where we need a stronger check like this. I'm happy to help with those if folks can test. Just let me know.

bgoglin:
We need to disable this check when NVML is built as a hwloc plugin, but I have to move lots of things inside configure before that's possible.

Unless I am missing something, the summary would be:
1) in your case without the NVIDIA driver:
BEFORE: builds with NVML, fails to run
AFTER: builds without NVML and runs
2) on platforms with login nodes without NVIDIA driver:
BEFORE: builds with NVML, runs on compute nodes, fails on login nodes
AFTER: builds without NVML, runs everywhere
3) on same platforms with plugins:
BEFORE: builds with NVML, runs everywhere
AFTER: same

I am not sure of the implications of (2). We've been telling people to use plugins for GPU backends for this exact problem but it's not clear they did.

I wonder if we should just disable NVML by default (and maybe even CUDA, NVIDIA OpenCL, and AMD ROCm).

jjhursey:
Yeah, that's correct. I'm also thinking about the (2) case. With the patch as it is now, NVML will be disabled in that case, and everything will run everywhere fine. The question is, what if they intend on building on the login node but running on the compute node and expect NVML.

I'm currently thinking we make the runtime check non-excluding of support, and only result in a warning. So if the program cannot run (`AC_RUN_IFELSE` fails), then we print a big warning message like the following but proceed to build/link in the support.

```
# WARNING: NVIDIA Management Library (NVML) was detected, but programs
# linked against it cannot run on this machine. This situation often indicates that
# this machine does not have the hardware necessary to run the library.
#
# Sometimes, the machine building hwloc differs from the machine where
# hwloc will be running. As such, this warning can be ignored if you know that the
# library is functional where you intend to run hwloc.
# However, if that is not the situation, you may want to disable NVML support
# in hwloc (with the --disable-nvml option) in configure.
```

What do you think about that solution?

bgoglin:
Sure, we should at least do this. My only question is where to put this message. It should likely be at the end of configure or people will miss it (except if we put a 30 second sleep or the need to hit the keyboard to continue :)). But the end of configure can already be huge (the pkg-config block in the middle isn't very common fortunately):
```
config.status: linking ../utils/lstopo/lstopo-android.c to tests/hwloc/ports/lstopo-android.c
config.status: executing depfiles commands
config.status: executing libtool commands
config.status: executing chmoding-scripts commands

************************************************************************
Could not detect/enable some features such as libxml2 and Cairo support
because pkg-config isn't available.
************************************************************************

-----------------------------------------------------------------------------
Hwloc optional build support status (more details can be found above):

Probe / display I/O devices: no
Graphical output:            yes (Windows)
XML input / output:          basic
Netloc functionality:        no
Plugin support:              no
-----------------------------------------------------------------------------
```

When NVML is enabled, it appears in this line:
```
Probe / display I/O devices: PCI(linux) LinuxIO OpenCL CUDA NVML
```
Maybe add something short saying that NVML might be broken?
```
Probe / display I/O devices: PCI(linux) LinuxIO OpenCL CUDA NVML(mis-installed?)
```

bgoglin:
Or add a NVML specific line in the summary table for this case.

jjhursey:
Yeah, that's a good idea. Let me take a pass at it and push up a revision in a little bit.

jjhursey:
I just re-pushed to this branch with the following change.

For a machine in configuration (1) and (2) it will now build with NVML and report:
```
**** NVML configuration
checking nvml.h usability... yes
checking nvml.h presence... yes
checking for nvml.h... yes
checking for nvmlInit in -lnvidia-ml... yes
checking whether a program linked with -lnvidia-ml can run... no
checking whether nvmlDeviceGetMaxPcieLinkGeneration is declared... yes
**** end of NVML configuration
...
config.status: executing chmoding-scripts commands

************************************************************************
WARNING: NVIDIA Management Library (NVML) was detected, but programs
linked against it cannot run on this machine. This situation often
indicates that this machine does not have the hardware necessary to run
the library.

Sometimes, the machine building hwloc differs from the machine where
hwloc will be running. As such, this warning can be ignored if you know
that the library is functional where you intend to run hwloc.
However, if that is not the situation, you may want to disable NVML
support in hwloc with the --disable-nvml option in configure.
************************************************************************

-----------------------------------------------------------------------------
Hwloc optional build support status (more details can be found above):

Probe / display I/O devices: PCI(linux) LinuxIO CUDA NVML
NVML support:                yes, but not functional on this machine (see warning above)
Graphical output:            no
XML input / output:          full
Netloc functionality:        yes (without scotch)
Plugin support:              no
-----------------------------------------------------------------------------
```

For a machine in configuration (3) it will continue to build with NVML and report:
```
**** NVML configuration
checking nvml.h usability... yes
checking nvml.h presence... yes
checking for nvml.h... yes
checking for nvmlInit in -lnvidia-ml... yes
checking whether a program linked with -lnvidia-ml can run... yes
checking whether nvmlDeviceGetMaxPcieLinkGeneration is declared... yes
**** end of NVML configuration
...
config.status: executing chmoding-scripts commands

-----------------------------------------------------------------------------
Hwloc optional build support status (more details can be found above):

Probe / display I/O devices: PCI(linux) LinuxIO CUDA NVML
NVML support:                yes
Graphical output:            no
XML input / output:          full
Netloc functionality:        yes (without scotch)
Plugin support:              no
-----------------------------------------------------------------------------
```

If they pass `--disable-nvml` then it looks like the following:
```
-----------------------------------------------------------------------------
Hwloc optional build support status (more details can be found above):

Probe / display I/O devices: PCI(linux) LinuxIO CUDA
NVML support:                no
Graphical output:            no
XML input / output:          full
Netloc functionality:        yes (without scotch)
Plugin support:              no
-----------------------------------------------------------------------------
```

bgoglin:
Thanks! I'll merge this later today and backport to 2.7

PR #572, opened 2023-03-14
Update .gitignore to include cross-compile libtool artifacts

ncorgan:
When configured with a non-native --host value, the libtool file will be prefixed by the host, which the current .gitignore misses.


bgoglin:
Do you need this in stable branches too? Or is git master enough?

PR #573, opened 2023-03-15
Fix name scheme when building with --target

### LINKED ISSUE #560: configure --with-target= fails to install (sed cannot find prefixed scripts)
### What version of hwloc are you using?

2.7.0

### Which operating system and hardware are you running on?

Linux <COMPUTER_NAME> 4.4.0-19041-Microsoft #2311-Microsoft Tue Nov 08 17:09:00 PST 2022 x86_64 x86_64 x86_64 GNU/Linux

Machine (16GB total) + Package L#0
  NUMANode L#0 (P#0 16GB)
  Core L#0 + PU L#0 (P#0)
  Core L#1 + PU L#1 (P#1)

My computer runs Windows 10 Enterprise (WSL), with an Intel Pentium G3258, 2*8GB Patriot Viper DDR3-1600 RAM, and built-in graphics.

### Details of the problem

I am trying to build Hwloc. `./configure --prefix="/root/MiOSR/toolchain/i686-elf/" --target=i686-elf` works, and so does `make -s`. But when I try to run `make -s check`, test-hwloc-compress-dir.sh fails to run.

### Additional information

I tried to fix a previous error that was related to `gunzip` and `tar` failing to unzip `test-hwloc-compress-dir.input.tar.gz` and `test-hwloc-compress-dir.output.tar.gz`. So, I tried to unzip them manually.

Here's the full `test-suite.log`:
```
=============================================
   hwloc 2.9.0: utils/hwloc/test-suite.log
=============================================

# TOTAL: 8
# PASS:  7
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test-hwloc-compress-dir.sh
================================

ls: cannot access '/tmp/foosawUN6/test-hwloc-compress-dir.input': No such file or directory
Compressed 0 new topologies (0 were already compressed)
Kept 0 new topologies non-compressed (0 were already non-compressed)
/usr/bin/diff: /tmp/foosawUN6/test-hwloc-compress-dir.output: No such file or directory
FAIL test-hwloc-compress-dir.sh (exit status: 2)
```

clementFoyer:
Building with --target may add an unexpected prefix to the binary names. This commit uses the same mecanism to take this prefix in account.

Resolves #560

### DIFF
--- a/utils/hwloc/Makefile.am
+++ b/utils/hwloc/Makefile.am
@@ -1,6 +1,7 @@
 # Copyright © 2009-2021 Inria.  All rights reserved.
 # Copyright © 2009-2012, 2014, 2016 Université Bordeaux
 # Copyright © 2009-2014 Cisco Systems, Inc.  All rights reserved.
+# Copyright © 2023      Université de Reims Champagne-Ardenne.  All rights reserved.
 #
 # See COPYING in top-level directory.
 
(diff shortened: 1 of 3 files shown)


bgoglin (commented):
Can't you define variables such as
```
hwloc_ls_name=`echo "hwloc-ls$(EXEEXT)" | $(SED) -e "$(transform)"`
```
on normal lines instead of inside the dependencies of hooks?

clementFoyer:
I did it inside the hook target to be sure to limit any unintended side effect, although it forces the repetition for both the install and the uninstall hooks. I guess we are pretty safe that automake does not use any `*_name` variable.

To be honest, I am more annoyed of seeing the part between back ticks being repeated...

Issue #577, opened 2023-04-14
archivemount'ed input topology changes working directory, breaking lstopo export to relative path


bgoglin:
When reading the topology from a tarball using archivemount in 2.9+, we "chdir" inside the mount point (so that we can umount it early but still access it later, this avoids mount leaks on crashes). However exporting the lstopo output then becomes relative to the mountpoint instead of the original working directory. Commit 8864ce07deee6c74eae51a998ed8073609cde3fb prevents from actually writing the output inside the archive, we now get an error, but we should actually fix things to export relatively to the original working directory.

One way would be to make all relative paths absolute right after parsing the command-line. Another (likely better) way would be to just revert the working directory after loading the topology. Or we need to avoid chdir but I don't see another way to keep the mountpoint available after umounting it.

Low priority since only occurs with archivemount and relative path output.

bgoglin:
Closed by #585

PR #585, opened 2023-05-03
Archivemount restore working dir

clementFoyer:
This PR saves the current working directory and restores it after the topology has been loaded.

Solves #577 by ensuring the current directory does not change (at least not from a side effect of loading a topology).

### DIFF
--- a/utils/hwloc/misc.h
+++ b/utils/hwloc/misc.h
@@ -3,6 +3,7 @@
  * Copyright © 2009-2023 Inria.  All rights reserved.
  * Copyright © 2009-2012 Université Bordeaux
  * Copyright © 2009-2011 Cisco Systems, Inc.  All rights reserved.
+ * Copyright © 2023 Université de Reims Champagne-Ardenne.  All rights reserved.
  * See COPYING in top-level directory.
  */
 
@@ -26,6 +27,7 @@
 #ifdef HAVE_DIRENT_H
 #include <dirent.h>
 #endif
+#include <fcntl.h>
 #include <assert.h>
 
 extern void usage(const char *name, FILE *where);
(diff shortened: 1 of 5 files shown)


utils/hwloc/misc.h
```diff
@@ -325,6 +335,19 @@ hwloc_utils_enable_input_format(struct hwloc_topology *topology, unsigned long f
     err = hwloc_utils_enable_input_format(topology, flags, subdir, &sub_input_format, verbose, callname);
     if (!err)
       *input_format = sub_input_format;
+    else {
```
clementFoyer, line 338: The 4 lines here depend on/are linked to PR #584

clementFoyer:
The issue with the build is the same as [there](https://github.com/netblue30/firejail/issues/2696), i.e., CentOS 6 does not declare O_PATH which is required to use `fchdir`.

bgoglin:
If you can make your code conditional with #ifdef O_PATH, I'd be fine with only having the issue on very old kernels. I am not even sure archivemount works on them anyway.

utils/hwloc/misc.h
```diff
@@ -325,6 +335,19 @@ hwloc_utils_enable_input_format(struct hwloc_topology *topology, unsigned long f
...
+    }
+
+    err = fchdir(oldworkdir);
+    if (err) {
+      perror("Restoring current working directory");
+      close(oldworkdir);
+      return EXIT_FAILURE;
+    }
+
+    close(oldworkdir);
```
bgoglin, line 350: Not sure that's enough. The goal is to keep a reference on the archivemount'ed path during topology_load(), which happens after this code. So you'd likely need to move fchdir+close after load(). That's a bit ugly but I guess we'd need to make oldworkdir a global variable initialized to -1, and that variable would need to be added to all users of "input_format" macros :/

bgoglin (commented):
Can you actually merge the double single-commit message into a single single-commit message ? :)

clementFoyer:
Sorry, I missed your last comment. I merged the two messages into a single one.

bgoglin:
Thanks.

PR #587, opened 2023-05-12
x86: build cpukinds from AMD CPUID leaf 0x80000026

bgoglin:
Use coretype to build "AMDEfficiency" and "AMDPerformance" kinds.

Use power efficiency rankings to build kinds with native efficiencies, or just make E=0 and P=1.


bgoglin:
@mkuron thanks to your CPUID dump in #690, I was able to verify that this code mostly works:
```
CPU kind #0 efficiency -1 cpuset 0x00ff0ff0
  CoreType = AMD-ECore
CPU kind #1 efficiency -1 cpuset 0x0000f00f
  CoreType = AMD-PCore
```
Does this CPU report EfficiencyRanking? I see 1 in EfficiencyRankingAvailable=1 in the first level but the ranking I read is 0 for all cores (that's why I have "efficiency -1" (unknown) above).

Also, I am not sure I understand why AsymmetricTopology=1 only in the second level, EfficiencyRankingAvailable=1 only in first level (that one is documented), and HeterogeneousCores=1 in first 3 levels but not in 4th.

hwloc/topology-x86.c
```diff
@@ -583,9 +585,44 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
...
+              /* EfficiencyRankingAvailable */
+              data->has_power_efficiency_ranking = 1;
+              /* "a core with a lower value has intrinsically better power,
+               *  but potentially lower performance potential vs cores with a higher value."
+               */
+              infos->power_efficiency_ranking = (ebx >> 16) & 0xff;
+            }
+
+            infos->hybridcoretype = (ebx >> 28) & 0xf; /* 0 = P, 1 = E */
+            infos->hybridnativemodel = (ebx >> 24) & 0xf; /* 0 = Zen4 */
```
superm1, line 624: both of these fields are only 4 bits a piece

hwloc/topology-x86.c
```diff
@@ -583,9 +585,44 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
...
+              data->is_hybrid = 1;
+            }
+
+            if (eax & 0x20000000) {
+              /* EfficiencyRankingAvailable */
+              data->has_power_efficiency_ranking = 1;
+              /* "a core with a lower value has intrinsically better power,
+               *  but potentially lower performance potential vs cores with a higher value."
+               */
+              infos->power_efficiency_ranking = (ebx >> 16) & 0xff;
```
superm1, line 620: this field is only 8 bits

hwloc/topology-x86.c
```diff
@@ -583,9 +585,44 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
                     id);
 	infos->apicid = apic_id;
 	infos->otherids[level] = UINT_MAX;
+if (leaf == 0x80000026 && apic_id == 0) {
+  printf("HybridCPU info at level %u\n", apic_type);
+  printf(" AsymmetricCores = %u\n", (eax & 0x80000000)>>31);
+  printf(" HeteroCoreTyopology  = %u\n", (eax & 0x40000000)>>30);
+  printf(" EffRankingAvailable = %u\n", (eax & 0x20000000)>>29);
+  printf(" CoreType = %u\n", (ebx >> 28) & 0xf);
```
superm1, line 593: 4 bits for this field

hwloc/topology-x86.c
```diff
@@ -583,9 +585,44 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
...
+  printf(" Model = %u\n", (ebx >> 24) & 0xf);
```
superm1, line 594: 4 bits for this field

hwloc/topology-x86.c
```diff
@@ -583,9 +585,44 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
...
+  printf(" EffRanking = %u\n", (ebx >> 16) & 0xff);
```
superm1, line 595: 8 bits for this one

bgoglin: Well, 0xf is 4 bits and 0xff is 8bits so all these look OK?

superm1: mm I guess so

superm1:
> Also, I am not sure I understand why AsymmetricTopology=1 only in the second level

It's a topology indicator only.  IE how they're grouped into larger units.  It's meant to be used in conjunction with bit 30 and 29.  IE if it's "1" then software is supposed to discover more details from 30 and 29.

> EfficiencyRankingAvailable=1 only in first level (that one is documented)

I think APM explains that already.

>  HeterogeneousCores=1 in first 3 levels but not in 4th

It's a single socket system.

bgoglin:
@superm1 Is the "NativeModelId" CPUID info supposed to be the same for all cores on hybrid CPUs?
The doc says "This value is used by software to further differentiate implementation specific software visible properties between the cores". I am not sure if that applies to cores of different chips, or different cores of the same chip.

I couldn't find the PPR for Strix Point. In the 19h family PPR, I see "Zen4" as the only NativeModelID possible value. Could we see some hybrid CPUs return something like "Zen5" and "Zen5c" for different cores? Or will they all return "Zen5" while CoreType is the only field distinguishing between Ecore and Pcore?

By the way, in case you missed the question above:
Does this CPU report EfficiencyRanking? I see 1 in EfficiencyRankingAvailable=1 in the first level but the ranking I read is 0 for all cores (that's why I have "efficiency -1" (unknown) above).

superm1:
Yeah the publication of the docs sometimes lags a bit from the launch, sorry about that.  I double checked though and  `0h` means `Zen5` for `NativeModelId` on Strix.  There is no differentiation of Zen5 vs Zen5c at this field.  You would tell that from `CoreType` when called with the cpuid is called with level `1h (Core)`

> Does this CPU report EfficiencyRanking? I see 1 in EfficiencyRankingAvailable=1 in the first level but the ranking I read is 0 for all cores (that's why I have "efficiency -1" (unknown) above).

I guess not; but I'm not sure why.  There are definitely rankings utilized from the CPPC MSRs.

PR #594, opened 2023-06-09
[WIP DNM] levelzero: use zesInit() when available

bgoglin:
Don't merge until a compute-runtime release supports zesInit(). Blacklist the releases with unsupported zesInit() in the error messages before merging this PR.

Refs #687

TODO just drop releases without zesInit() and abort if the backend if zesInit() fails
TODO check whether zesInit() must be called in a constructor before any other zeInit() or just in the hwloc backend as usual.
TODO requires oneapi-src/level-zero#153


saik-intel:
@bgoglin  zesInit () is fully supported avaliable , can you start using zesInit()

bgoglin:
Obsolete, dropped in favor of #695

PR #596, opened 2023-06-13
Clean utils args

clementFoyer:
Mostly code refactoring.

Unify the definition of options to increase code readability. Reorder the option checks to enforce topology options to be parsed first. Unmix using tabs and spaces (excepted for cases when the line is cut due to the function's arguments). Enforce using NULL character instead of simply `0`.

### DIFF
--- a/utils/hwloc/hwloc-annotate.c
+++ b/utils/hwloc/hwloc-annotate.c
@@ -14,32 +14,32 @@
 
 void usage(const char *callname __hwloc_attribute_unused, FILE *where)
 {
-	fprintf(where, "Usage: hwloc-annotate [options] <input.xml> <output.xml> -- <location1> <location2> ... -- <annotation>\n");
-	fprintf(where, "Usage: hwloc-annotate [options] <input.xml> <output.xml> <location> <annotation>\n");
-	fprintf(where, "  <location> may be:\n");
-	fprintf(where, "    all, root, <type>:<logicalindex>, <type>:all\n");
-	fprintf(where, "  <annotation> may be:\n");
-	fprintf(where, "    info <name> <value>\n");
-	fprintf(where, "    subtype <subtype>\n");
-	fprintf(where, "    size <memory or cache size>\n");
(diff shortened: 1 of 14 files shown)


utils/hwloc/hwloc-annotate.c
```diff
@@ -599,24 +603,24 @@ int main(int argc, char *argv[])
...
+				sizevalue <<= 20;
+			else if (!strcasecmp(end, "GB"))
+				sizevalue *= 1024ULL*1024ULL*1024ULL;
+			else if (!strcasecmp(end, "GiB"))
+				sizevalue <<= 30;
+			else if (!strcasecmp(end, "TB"))
+				sizevalue *= 1024ULL*1024ULL*1024ULL*1024ULL;
+			else if (!strcasecmp(end, "TiB"))
+				sizevalue <<= 40;
+		}
```
clementFoyer, line 623: It seems that the treatment is the same whether the extension is `[kMGT]B` or `[kMGT]iB`. Isn't shifting left 10 times and multiplying by `1024` the same ?

bgoglin: Looks like commit 4d9a9843a5616fc702575c4701a8c3cba044757f failed to fix [kMGT]B contrary to what the commit message said :)

bgoglin:
First 2 commits are OK.  3rd one will require 3 months of careful review and mixes several things. I don't like the 4th commit because it makes git blame a bit harder. I usually reindent only when the modify the code nearby.

bgoglin:
I pushed the first two commits and fixed the hwloc-annotate bug with units.

clementFoyer:
I can split it by files and/or by changes (although the changes that are not just the reordering of the options are usually close to it).

No problem for the fourth commit. I think you may have told me in the past, but I had forgotten it. Anyhow, I actually put in a separated commit because I knew it was more likely to be refused.

bgoglin:
I pushed the first commit, I'll try to review the last one later today.

PR #600, opened 2023-06-25
windows: Fix type of hostname_size parameter

evan-charmworks:
Avoids warning: passing argument 2 of 'GetComputerNameA' from incompatible pointer type [-Wincompatible-pointer-types]


bgoglin:
Applied, thanks!

PR #608, opened 2023-07-31
Add memory properties

xiongzubiao:
- CXL
  - Serial number
- DRAM:
  - Form factor
  - Type
  - Size
  - Rank

### DIFF
--- a/hwloc/topology-linux.c
+++ b/hwloc/topology-linux.c
@@ -6896,6 +6896,8 @@ struct hwloc_firmware_dmi_mem_device_header {
   unsigned char serial_str_num;
   unsigned char asset_tag_str_num;
   unsigned char part_num_str_num;
+  unsigned char attributes;
+  unsigned char extended_size[4];
   /* don't include the following fields since we don't need them,
    * some old implementations may miss them.
    */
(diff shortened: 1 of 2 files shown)


xiongzubiao:
@bgoglin I found these useful properties are actually easy to get based on the current source code. Can you please kindly have a look? BTW, I am not quite sure how to look at the Jenkins failure. Can you please shed some light?

bgoglin:
Hello. When you click on "Details" above, you go to the build on that version of the PR that fails. Click on "pipeline steps" on the left, you'll find that several steps failed. Click on the terminal to see the log. Unfortunately, in this case, we just see "FAIL: 32em64t-2n8c+dax+nvme+mic+dimms.test".

 I have to go on the node to get more details. There, I see that MemoryModule size got parsed incorrectly, it's set to 0 instead of 4GB. Given that things failed on 32bits Linux only, I guess you have some 32/64 issue in the parsing of the size. This seems to work better:
```
-  snprintf(buffer, len, "%lu", memory_size);
+  snprintf(buffer, len, "%llu", (unsigned long long) memory_size);
```

For additional DIMM attributes, I think I wrote the code for spec 2.3 but your new ones are for spec 2.6 only. We may want to check the header size before using them? And also I'd move things like `*(uint16_t *)(&header->size), *(uint32_t *)(&header->extended_size)` inside your new dmi_memory_foo() functions (and check the header length there).
 
Regarding your commit, can you add a Signed-off-by line and use /* */ comment instead of // ? And remove some trailing whitespaces.

xiongzubiao:
Thanks for the pointers! The bug has been fixed now.

Also added the header length check for the two 2.6+ fields and the signed-off-by line as suggested. Fixed the comment style and whitespaces.

bgoglin:
Thanks, I'll review carefully tomorrow, test on several servers, and likely merge this in master and v2.x (for the future 2.10 release). At least it seems to work fine on my laptop :)
Also I'll add a little bit of doc in doc/hwloc/doxy, but I can't find what the "rank" means, can you explain?
By the way, we should likely ignore the rank when it's 0, the spec says 0 means unknown.

xiongzubiao:
Thanks for the thorough review and tests!

Your intended changes all sound good to me. Looking forward to the merge! :-)

bgoglin:
Can you tell me what the "Rank" is?

xiongzubiao:
> Can you tell me what the "Rank" is?

I am no expert on this :-). In my understanding, it is a block of 64-bit wide data area. It could be composed of multiple physical chips, depending on the width of each chip. The modern DIMMs are trying to pack multiple ranks to  increase the memory density. However, there could be a limitation on how many ranks the controller can support. Regarding the performance, it sounds like multiple ranks should be slower than single rank because of sharing the data bus. But for certain scenarios the rank interleaving might make it perform better.

bgoglin:
Ah ok, thanks, it's the number of ranks within the DIMM (thus 0=unknown makes sense), I was wondering if it was the rank of the DIMM within a channel or something like that.

bgoglin:
Pushed in master (abfd6136e0fa4c0d7dff9b2260e4247ef27bffa2) and in v2.X branch (it will be in hwloc 2.10, likely within 2 months). Thanks!

PR #654, opened 2024-03-13
lstopo/x11: fix Unknown icon shown on GNOME

cjmayo:
Set WM_CLASS to help GNOME Shell associate an instance of lstopo with lstopo.desktop.

---

To reproduce start `lstopo` and then press Alt-tab.

It can be seen on Debian Linux 12 with its hwloc package.

Commit tested against hwloc 2.10.0 on Gentoo Linux with GNOME 45.

### DIFF
--- a/utils/lstopo/lstopo-cairo.c
+++ b/utils/lstopo/lstopo-cairo.c
@@ -390,7 +390,7 @@ output_x11(struct lstopo_output *loutput, const char *dummy __hwloc_attribute_un
   XSetIconName(dpy, top, "lstopo");
   class_hint = XAllocClassHint();
   if (class_hint) {
-    class_hint->res_name = class_hint->res_class = "lstopo";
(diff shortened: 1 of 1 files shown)


bgoglin:
I don't see any change in "alt-tab" under XFCE, any idea why?

Also, what should we do about this warning:
```
warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  393 |     class_hint->res_name = class_hint->res_class = "lstopo";
```

bgoglin:
I found a machine with Gnome, I don't see a different icon in alt-tab, but it says "Hardware Locality" instead of "unknown" under it. On my XFCE, it says "lstopo" instead of "...".

bgoglin:
I squashed your commit and pushed them to master, will backport to 2.11 for next release, thanks!

bgoglin:
I am posting 2.11rc1 right now with this fix.

PR #656, opened 2024-03-14
windows: fix dll filenames vs text encoding

### LINKED ISSUE #655: LoadLibrary("kernel32.dll") should be LoadLibrary(TEXT("kernel32.dll")) in topology-windows.c
`LoadLibrary` automatically switches between `LoadLibraryW` and `LoadLibraryA` when UNICODE is enabled or disabled but the given string `"kernel32.dll"` does not and the call fails. The `TEXT` macro fixes this.

`GetModuleHandle` has the same issue.

bgoglin:
Thanks to MKoep for the report.

Closes #655


bgoglin:
Merged in b4ce80732bb9abd699f68229db3c489669777351

PR #671, opened 2024-06-27
add the systemd output format

npf:
Add format expected by the systemd dbus api.


npf:
Replaced by https://github.com/open-mpi/hwloc/pull/684, in order to not impact the hwloc API, as discussed.

Issue #676, opened 2024-07-18
build hwloc 2.11.1 on old suse 9: undefined reference to `hwloc_accessat' 


zhchgbj:
### What version of hwloc are you using?
2.11.1

### Which operating system and hardware are you running on?
I'm using my old box of suse 9 to build hwloc 2.11.1. 

```
ff75ff964fbd:/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64 # uname -a
Linux ff75ff964fbd 5.14.0-427.22.1.el9_4.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Jun 10 09:23:36 EDT 2024 x86_64 x86_64 x86_64 GNU/Linux
ff75ff964fbd:/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64 # cat /etc/*release
LSB_VERSION="core-2.0-noarch:core-3.0-noarch:core-2.0-x86_64:core-3.0-x86_64"
SUSE LINUX Enterprise Server 9 (x86_64)
VERSION = 9
PATCHLEVEL = 4

```
### Details of the problem

When run `make`, it report `hwloc_accessat` not defined.

```
  CC       topology-linux.lo                                                                                                                                
  CC       topology-hardwired.lo                                                                                                                            
  CC       topology-x86.lo                                                                                                                                  
topology-linux.c: In function `hwloc_find_linux_cgroup_mntpnt':                                                                                             
topology-linux.c:2290: warning: implicit declaration of function `hwloc_accessat'                                                                           
  CCLD     libhwloc.la                                                                                                                                      
make[1]: Leaving directory `/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64/hwloc'                                                         
Making all in utils                                                                                                                                         
make[1]: Entering directory `/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64/utils'                                                        
Making all in hwloc                                                                                                                                         
make[2]: Entering directory `/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64/utils/hwloc'                                                  
Making all in .                                                                                                                                             
make[3]: Entering directory `/data/code/3rd/hwloc/build/hwloc-2.11.1/linux2.6-glibc2.3-x86_64/utils/hwloc'                                                  
[...]

PR #677, opened 2024-07-19
fix#676 use hwloc_access instead of hwloc_accessat

zhchgbj:
Fix https://github.com/open-mpi/hwloc/issues/676:

`HAVE_OPENAT` should not defined on old suse 9.
Change `hwloc_accessat` to inline version `hwloc_access` as it  preserves compiler parameter checking.


bgoglin:
Thanks, I am merging and backporting to affected releases (2.11 iirc).

PR #695, opened 2024-11-08
L0: switch to zesInit()

bgoglin:
- only support implementations with zesDriverGetDeviceByUuidExp()
- the above means we may assume zesInit() is available and works
- remove the constructor that forces ZES_ENABLE_SYSMAN in the environment (FINALLY!!!)
- use different handles for ZE and ZES devices during discovery
- cleanup/update helpers and tests accordingly

Still needs to be tested on a machine with subdevices (PVC), but I don't have any such machine with a recent-enough L0 implementation.

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -1410,10 +1410,9 @@ return clGetDeviceIDs(0, 0, 0, NULL, NULL);
       echo
       echo "**** LevelZero configuration"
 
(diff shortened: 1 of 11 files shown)


TApplencourt:
>Still needs to be tested on a machine with subdevices (PVC), but I don't have any such machine with a recent-enough L0 implementation.

BTW,  when I tests it was  on a PVC

Thanks a lot!

PR #714, opened 2025-04-16
misc: reactivate GNUC conditional

tacerus:
This avoids use of the custom hwloc_flsl_manual() implementation if a recent GCC is in use.
It is not clear why the erroneous underscores, with which the macro does not seem to be defined anywhere, were introduced, as the commit introducing the line only references an SVN import. Older source releases however do not contain the change.

This was discovered as hwloc_flsl_manual() returns zero and as a result a division by zero exception with recent versions of GCC on certain CPU configurations, whereas the native functionality behaves correctly on both tested versions.
The manual implementation being defective is technically still an issue, however the logic under the __GNUC__ conditional is deemed to be preferable in any case.

Fixes: 3ac17a230d2a ("topo_cpuset_last")
Link: https://bugzilla.opensuse.org/show_bug.cgi?id=1236038
Signed-off-by: Georg Pfuetzenreuter <mail@georg-pfuetzenreuter.net>

### DIFF
--- a/include/private/misc.h
+++ b/include/private/misc.h
@@ -186,7 +186,7 @@ hwloc_ffsl_from_ffs32(unsigned long x)
 /**
  * flsl helpers.
  */
-#ifdef __GNUC_____
+#ifdef __GNUC__
 
 #  if (__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 4))
 #    define hwloc_flsl(x) ((x) ? (8*sizeof(long) - __builtin_clzl(x)) : 0)


bgoglin:
Hmmm thanks. I am not 100% confortablke with changing with very old code but I guess I'll merge it anyway. But it'd be great to open the issue if the manual implementation is buggy, even if we don't fix it.

@sthibaul Can you confirm that this __GNUC__ with way too many underscores is likely not intentional?

bgoglin:
@tacerus Could you post some details about failures in the manual case? I'd like to see if I can add some regression tests comparing the manual and GNUC versions.

sthibaul:
> @sthibaul Can you confirm that this **GNUC** with way too many underscores is likely not intentional?

Sorry I missed that notification. It's most probably not intentional indeed.

tacerus:
Hi @bgoglin, thanks for the reply.
I had asked the colleague who investigated the manual logic more than me to open an issue as per your request, but I think he's still away and did not get to it yet.

This is what he wrote last:

> Basically: legacy_max_log_proc = 1 << hwloc_flsl(((ebx >> 16) & 0xff) - 1); in this case (ebx >> 16) & 0xff is zero, and then becomes 0xffff (due to -1), before fed into hwloc_flsl(). With gcc-13 this would return 1 while with gcc-14 returns 0.

I think he'll be happy to share more details when back.

bgoglin:
Hmm, what's returning 0 or 1 above? flsl(0xffff) is supposed to be 16 (and that's what I get from GCC 14.2, 13.3 and 12.4 with both our manual implementation and GCC intrinsics.
Any chance you get the exact GCC 13 version?

By the way, are you running in a virtualized environment? (ebx>>16)&0xff isn't supposed to 0, but it's a legacy field so maybe some hypervisors don't implement it.

tacerus:
Our gcc 13 is at

```
13.3.1 20250313 [revision 4ef1d8c84faeebffeb0cc01ee22e891b41e5c4e0]
```

though I'm not sure if that's still the same exact version the problem happened with, there were updates in the meanwhile which I did not test.

Yes, the issue was only observed in KVM guests. If useful, the QEMU command line and cpuid information of one of the affected VMs can be found in the attachments of https://bugzilla.opensuse.org/show_bug.cgi?id=1236038.

sthibaul:
> > Basically: legacy_max_log_proc = 1 << hwloc_flsl(((ebx >> 16) & 0xff) - 1); in this case (ebx >> 16) & 0xff is zero,

Which is the original bug actually: the code assumes to get something different from zero, to be able to compute a log out of it. The if before should check against it.

> and then becomes 0xffff (due to -1), before fed into hwloc_flsl(). With gcc-13 this would return 1 while with gcc-14 returns 0.

I guess the problem is not really the manual implementation itself (which does, as expected, return 32 when given -1), but the interaction in the resulting 1 << 32 computation, which is undefined behavior, and thus depending on whether using hwloc_flsl_manual or __builtin_clzl, it may get different behavior.

So I believe that while using the builtin will be better indeed, the calling code should care for the zero-case ahead of this.

bgoglin:
Indeed in Qemu, CPUID.0x1.ebx[16:23] seems to be 0x00 on my VMs too. I'll stop calling flsl() in this case.

bgoglin:
More details:
Passing ht=on in the qemu -cpu parameter sets bit CPUID.1.EDX[28], that's why hwloc enters the if that leads to flsl(). However Qemu puts 0 in CPUID.1.EBX[16-23] because  -smp 32,maxcpus=48,sockets=48,cores=1,threads=1 says each CPU is single threaded. So flsl() is called on a buggy value.

This configuration seems invalid on both AMD and Intel. On recent AMD CPUs (not sure about older ones), CPUID.1.EBX[16-23] is never 0. On Intel, it may only be 0 if bit CPUID.1.EDX[28] is also unset (very old CPU).

Removing ht=on from qemu -cpu should avoid the issue. In theory, it could be a qemu bug but it could also be considered a misconfiguration from the user.

Anyway I'll document the fix better now that I understand.

tacerus:
Thanks for the details! Note that this QEMU configuration is in my case generated by Libvirt, which does not seem to have an option to toggle `ht=on` specifically. So if it's considered a misconfiguration, I think it's also an issue in the Libvirt abstraction logic.

bgoglin:
I pushed your fix and the x86 CPUID one to master, will backport to 2.12 and release 2.12.1 soon. Thanks!

sthibaul:
> The manual implementation being defective is technically still an issue

No, it is *not* defective. It does not return 0, but 32, as expected. And then 1<<32 is undefined, and that's where everything comes from, anything can happen from that.

sthibaul:
> > The manual implementation being defective is technically still an issue
> 
> No, it is _not_ defective. It does not return 0, but 32, as expected. And then 1<<32 is undefined, and that's where everything comes from, anything can happen from that.

Just to make it clear: my understanding of `With gcc-13 this would return 1 while with gcc-14 returns 0.` is that it was not referring to the value returned by hwloc_flsl_manual (there is no way it could return 0 since the parameter is not 0, 1 would be even less probable), but to the value stored in legacy_max_log_proc (i.e. the undefined 1 << 32, which can be 0 or 1 depending on what the assembler backend or whatnot sees most convenient to utter).

tacerus:
Thank you @bgoglin for the proper solution and @sthibaul for elaborating.

PR #715, opened 2025-04-17
hwloc_type_sscanf: fix test when host has several levels with same type

sthibaul:
E.g. with

HWLOC_SYNTHETIC="group:2 group:2 pu:2" ./tests/hwloc/hwloc_type_sscanf

The presence of the two group levels makes the check for level depth fail for the hwloc_obj_type_string case which does not expose attributes which allow to distinguish them.


sthibaul:
I have backported it to v2.x

PR #724, opened 2025-07-21
lstopo-ascii: Use a pager

sthibaul:
Otherwise the output essentially always wraps around, making it unreadable.

### DIFF
--- a/config/hwloc.m4
+++ b/config/hwloc.m4
@@ -1,7 +1,7 @@
 dnl -*- Autoconf -*-
 dnl
 dnl Copyright © 2009-2025 Inria.  All rights reserved.
-dnl Copyright © 2009-2012, 2015-2017, 2020, 2023 Université Bordeaux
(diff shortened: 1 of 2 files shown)


bgoglin:
The failure on Debian 9 is when building with -m32 hence configure fails to find curses. In this case, we get a textual output without colors, but still with black and white boxes.
```
checking termcap support using ncurses and ... no
checking termcap support using ncurses and -lncursesw... no
checking termcap support using ncurses and -lncurses... no
checking termcap support using ncurses and -ltinfo... no
checking termcap support using ncurses and -ltermcap... no
checking termcap support using curses and ... no
checking termcap support using curses and -lcursesw... no
checking termcap support using curses and -lcurses... no
checking termcap support using curses and -ltinfo... no
checking termcap support using curses and -ltermcap... no
```
The full log is at https://ci.inria.fr/hwloc/job/basic/view/change-requests/job/PR-724/1/execution/node/175/log/

Issue #728, opened 2025-07-28
hwloc_internal_distances_dup leaves grouping info uninitialized


HadrienG2:
In a topology where distances are enabled, at the time where the topology is loaded, [`hwloc_internal_distances_prepare()`](https://github.com/open-mpi/hwloc/blob/59f7156758591743d4612db205d9c75ebdacddf8/hwloc/distances.c#L63) sets up some topology fields whose name starts with `grouping`, which together define how objects should be grouped  when new distances are added with the `HWLOC_DISTANCES_ADD_FLAG_GROUP` flag on.

If that topology is later duplicated, the associated [`hwloc_internal_distances_dup()`](https://github.com/open-mpi/hwloc/blob/59f7156758591743d4612db205d9c75ebdacddf8/hwloc/distances.c#L186) hook does not duplicate this information. Therefore, fields of the new topology object whose name starts with `grouping` will be left in an uninitialized state with unpredictable values.

This will cause all sorts of mayhem if one later attempts to add distances with the `HWLOC_DISTANCES_ADD_FLAG_GROUP` flag. In particular...

- `grouping` will have an unpredictable value that is unlikely to be 0, and thus grouping will be considered enabled even if it should be disabled by type filter or `HWLOC_GROUPING` environment.
- `grouping_verbose` will have an unpredictable value that is unlikely to be 0, and thus verbose grouping logs will be printed even if they were not enabled.
- `grouping_nbaccuracies` will have an unpredictable value that is unlikely to be in bounds of the associated `grouping_accuracies` array (which otherwises contains garbage), and thus any attempt to access that array (such as those made by the aforementioned verbose logs) will result in an arbitrary deep out-of-bounds memory read, hopefully ending with a segfault.

I think resolving this is just a matter of copying all members whose name starts with `grouping` from `old` to `new` in `hwloc_internal_distances_dup()`. But since I'm not sure if I fully understand what these members all do, I just wanted to check with you that a dumb copy is indeed appropriate.

PR #729, opened 2025-07-28
Duplicate distance grouping info in hwloc_internal_distances_dup

HadrienG2:
This implements the dump "copy all the grouping info" fix to the bug described in #728 .


HadrienG2:
Thanks for the suggestions! Applied them in the latest commit.

bgoglin:
Yes, we have lots of such issues with both windows slave, right now. I'll rerun that test.

Issue #726, opened 2025-07-25
Build Failure on Windows ARM64 (MSVC)


Willaaaaaaa:
### Description
Hello,

I am in the process of packaging your library to [xmake](https://xmake.io/)'s package manager to make it easily accessible to more developers.
During this process, the automated CI detected build failures when compiling for the **Windows** on **ARM64** with **MSVC** toolchain. 

### Version
2.12.1 (building from the source code)

### OS and Arch
The build is running on Github Actions, which reports its environment as:
- Microsoft Windows 11 Desktop by Arm Limited (ARM64)
- Microsoft Windows Server 2022 (ARM64)

### Details of the problem

* What happened?
  The library fails to build from source using the CMake build system provided in `./contrib/windows-cmake`.
* How did it fail?
  It fails during compilation. Here is the key error logs from ci:
  1. Microsoft Windows 11 Desktop by Arm Limited (ARM64)
  > cpuid-x86.h(88): error C2143: syntax error: missing ')' before ':'

  the whole log see [here](https://github.com/xmake-io/xmake-repo/actions/runs/16503456918/job/46667947337?pr=7733#step:5:773)

  2. Microsoft Windows Server 2022 (ARM64)
  > cpuid-x86.h(103): fatal error C1189: #error:  unknown architecture

  the whole log see [here](https://github.com/xmake-io/xmake-repo/actions/runs/16503456920/job/46667947651?pr=7733#step:5:689)

### Root Cause Analysis
The root cause appears to be in `contrib/windows-cmake/CMakeLists.txt`.
1. Unlike your automake, the `topology-x86.c` source file is unconditionally added to the `hwloc` library target sources when using CMake. This file, along with its included header `cpuid-x86.h`, contains assembly that are not recognized by the MSVC compiler with ARM64 arch.
https://github.com/open-mpi/hwloc/blob/a397c428e80bd97425de833341c6fac646409839/hwloc/Makefile.am#L178-L180
https://github.com/open-mpi/hwloc/blob/a397c428e80bd97425de833341c6fac646409839/contrib/windows-cmake/CMakeLists.txt#L125
https://github.com/open-mpi/hwloc/blob/a397c428e80bd97425de833341c6fac646409839/contrib/windows-cmake/CMakeLists.txt#L53-L61
2. The CMake script seems to fail in detecting the right arch.

Willaaaaaaa:
[...]

PR #732, opened 2025-07-29
CMake: add proper win-arm64 architecture detection

Willaaaaaaa:
#726 

## Main changes:
in `contrib/windows-cmake/CMakeLists.txt`

## Problem:
The existing code relies on `CMAKE_SYSTEM_PROCESSOR` to distinguish the target architecture, which is not reliable on Windows (see CMake discussions [here](https://gitlab.kitware.com/cmake/cmake/-/issues/15170)). This can cause compilation failures, because source file(`topology-x86.c`) will be wrongly included in ARM64 builds.

## Solution:
Inspired by the discussions, [CMAKE_C_COMPILER_ARCHITECTURE_ID](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ARCHITECTURE_ID.html) is here to replace `CMAKE_SYSTEM_PROCESSOR`. And then, conditionally add `topology-x86.c` to the library.

This should enable successful native & cross compiling.

### DIFF
--- a/contrib/windows-cmake/CMakeLists.txt
+++ b/contrib/windows-cmake/CMakeLists.txt
@@ -50,15 +50,20 @@ set(SIZEOF_VOID_P ${CMAKE_SIZEOF_VOID_P})
 set(HWLOC_X86_32_ARCH)
 set(HWLOC_X86_64_ARCH)
 set(HWLOC_HAVE_X86_CPUID 1)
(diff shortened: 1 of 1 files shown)


bgoglin:
And if you give me your name, I'll thank you in the NEWS file :)

Willaaaaaaa:
Thank you, it would be an honor! You can credit me as Chiwen Hui.

PR #740, opened 2025-09-09
Optional XML output from doxygen.

trivialfis:
To generate XML output using doxygen, specify the environment variable `HWLOC_DOXYGEN_GENERATE_XML=YES` when running doxygen:

```sh
HWLOC_DOXYGEN_GENERATE_XML=YES doxygen ./doxygen.cfg
```

If the environment variable is not set, doxygen will use the default `NO`.


bgoglin:
Oh I didn't know it was possible to use environment variables in the doxygen config.

PR #757, opened 2026-01-12
[WIP] linux: try mount points with statmount

kphuphanwoe:
I'm trying implement function find cgtype via linux `listmount`/`statmount` APIs. These syscall are not supported in linux kernel older than 6.8. btw I'm using fedora 42 which does support them

### DIFF
--- a/hwloc/topology-linux.c
+++ b/hwloc/topology-linux.c
@@ -1,7 +1,7 @@
 /*
  * SPDX-License-Identifier: BSD-3-Clause
  * Copyright © 2009 CNRS
- * Copyright © 2009-2025 Inria.  All rights reserved.
+ * Copyright © 2009-2026 Inria.  All rights reserved.
  * Copyright © 2009-2013, 2015, 2020 Université Bordeaux
  * Copyright © 2009-2018 Cisco Systems, Inc.  All rights reserved.
  * Copyright © 2015 Intel, Inc.  All rights reserved.
@@ -40,6 +40,7 @@
 #include <mntent.h>
 #include <stddef.h>
 #include <endian.h>
+#include <sys/mount.h>
 
 struct hwloc_linux_backend_data_s {
   char *root_path; /* NULL if unused */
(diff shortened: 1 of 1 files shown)


bgoglin:
Hello. Thanks for working on this. I understand this is still WIP. Two comments:
* Don't increase the Inria copyright to 2026 in the headers, that's my employer. You may add your own copyright line for 2026, either with your own name if you're doing this on your free time, or your employer if they agree. It's not clear if those copyright headers are useful nowadays. git blame has all the information line by line anyway, so making sure you are properly identified in commits or even signing commits with "Signed-off-by" is often used instead these days.
* I am surprised that you need to define listmount/statmount syscalls. We do that for NUMA syscalls because those syscalls are in numactl instead of glibc and we don't want to depend on numactl devel headers (not always installed). However the situation here looks different. It would be nice to understand what glibc is going to do. Are they going to expose another API on top of the syscalls? Just reimplement the old one?

kphuphanwoe:
I have also been considering that defining the syscall numbers may not be very useful, and we may delete this part later.

The reason we define our own functions is that glibc does not yet expose any binding for listmount/statmount, However syscall numbers are currently available.

The copyright was updated because pre-commit failed during commit.

bgoglin:
> The copyright was updated because pre-commit failed during commit.

Set these in your environment so that it updates with the correct lines:
HWLOC_COPYRIGHT_FORMAL_NAME (full line including things like All rights reserved etc)
HWLOC_COPYRIGHT_SEARCH_NAME (single word that we'll use for when searching your line for updating years)

For me, my employers wants this:
HWLOC_COPYRIGHT_FORMAL_NAME="Inria.  All rights reserved."
HWLOC_COPYRIGHT_SEARCH_NAME=Inria

If you don't care, remove the pre-commit hook, or commit with --no-verify.

bgoglin:
> The reason we define our own functions is that glibc does not yet expose any binding for listmount/statmount, However syscall numbers are currently available.

Do you know more about this "yet" ? Is it missing because the glibc API isn't ready yet? Or do they want to keep the syscall internal instead of exposing it directly? (we've seen some cases like that in the past)

kphuphanwoe:
I don't have more info here, "yet" I just meant glibc doesn't currently expose a public API for these syscalls

PR #759, opened 2026-01-22
lstopo/x11: add desktop translations

tblume:
add international translations to lstopo.desktop


bgoglin:
Hello. The name part is awkward according to my Brazilian colleague at least. And I am not sure we should have a different name for each language. The project is called "hardware locality" and "lstopo" in all languages anyway. Comment and keywords are better, but did you actually verify that those are the standard keywords for all these languages? Or did you just translate on the web without understanding/verifying the output?

PR #777, opened 2026-05-22
xml/diff: free partial diff list when import fails

dxbjavid:
hwloc__xml_import_diff() returns -1 when find_child errors out or a non-"diff" child is hit, but if some <diff> entries were already accepted, firstdiff (and the strdup'd name/oldvalue/newvalue inside each entry) is leaked: the caller sees *firstdiffp still NULL and has no way to free it.

Reproduced with hwloc_topology_diff_load_xmlbuffer() on a `<topologydiff>` containing two valid `<diff type="0" obj_attr_type="2" .../>` entries followed by a `<bogus/>` element. `leaks --atExit` reports 8 leaks / 224 bytes before the patch, 0 after.


bgoglin:
Thanks!

PR #778, opened 2026-05-23
xml/nolibxml: reject zero or negative input buffer length

dxbjavid:
hwloc_nolibxml_backend_init() and hwloc_nolibxml_import_diff() do `buffer[xmlbuflen-1] = '\0'` after `malloc(xmlbuflen)` of the caller-supplied buffer. When hwloc_topology_set_xmlbuffer() or hwloc_topology_diff_load_xmlbuffer() is called with `size <= 0`, malloc returns a non-NULL 0-byte chunk on macOS/glibc and the write lands one byte before it. Reject with EINVAL before the allocation.


bgoglin:
Thanks

Issue #776, opened 2026-05-19
Graphics locality support (OpenGL, Vulkan, anything else?)


bgoglin:
GL support must be disabled in some cases to avoid hangs (#517). It's not clear if it's still useful, we may remove that backend at some point.

If somebody cares about GPU locality, we'll need to look at the API they use and find something better than the current GL implementation.

Last time I checked, Vulkan had vkGetPhysicalDeviceProperties2() but it didn't work. Sample code below.
```
#include <vulkan/vulkan.h>
#include <stdio.h>
#include <stdlib.h>

int main()
{
  VkInstanceCreateInfo instanceCreateInfo = {
    VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO,
    NULL,
    0, NULL,
    0, NULL,
    0, NULL,
  };
  VkInstance inst;
  VkPhysicalDevice *phys;
  uint32_t n, i, j;

  vkCreateInstance(&instanceCreateInfo, NULL, &inst);

  n = 0;
  vkEnumeratePhysicalDevices(inst, &n, NULL);
  printf("found %u devices\n", n);

  if (!n)
    goto out;

  phys = malloc(n *sizeof(*phys));
  if (!phys)
    goto out;
  vkEnumeratePhysicalDevices(inst, &n, phys);

  for(i=0; i<n; i++) {
    VkPhysicalDeviceProperties prop;
    vkGetPhysicalDeviceProperties(phys[i], &prop);
    printf("type %d vendor %x device %x name %s uuid ", prop.deviceType, prop.vendorID, prop.deviceID, prop.deviceName);
    for(j=0; j<VK_UUID_SIZE; j++)
      printf("%hhx", prop.pipelineCacheUUID[j]);
    printf("\n");
  }

  for(i=0; i<n; i++) {
    VkPhysicalDevicePCIBusInfoPropertiesEXT bi = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PCI_BUS_INFO_PROPERTIES_EXT};
    VkPhysicalDeviceProperties2 prop = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2, &bi};
    vkGetPhysicalDeviceProperties2(phys[i], &prop);
    printf("%04x:%02x:%02x.%01x\n", bi.pciDomain, bi.pciBus, bi.pciDevice, bi.pciFunction);
  }

 out:
  vkDestroyInstance(inst, NULL);
  return 0;
[...]

PR #779, opened 2026-05-25
vulkan gpu support

ct-clmsn:
initial inclusion of vulkan GPU (cross vendor) support

see #776


ct-clmsn:
@bgoglin not sure what the issue was with CI but feel free to consider this code a building block to inform hwloc vulkan support. I have projects that use this hwloc vulkan patch currently and I'm completely unattached to this specific implementation. Getting any official vulkan support in hwloc would be greatly appreciated. I just hope this PR expedites a feature and solution.

bgoglin:
Hello. Thanks a lot for all this. We've been talking about this very recently in #776 for my limited knowledge of Vulkan prevented me from getting the PCI locality extension working.

CI fails because you need to add this:
```
--- i/include/hwloc/rename.h
+++ w/include/hwloc/rename.h
@@ -796,6 +796,7 @@ extern "C" {
 #define hwloc_levelzero_component HWLOC_NAME(levelzero_component)
 #define hwloc_nvml_component HWLOC_NAME(nvml_component)
 #define hwloc_rsmi_component HWLOC_NAME(rsmi_component)
+#define hwloc_vulkan_component HWLOC_NAME(vulkan_component)
 #define hwloc_opencl_component HWLOC_NAME(opencl_component)
 #define hwloc_pci_component HWLOC_NAME(pci_component)
 
```
Your Cmake stuff splits the current code and then adds Vulkan stuff (and reindents some lines). I'll like to distinguish those steps and just include Vulkan additions without moving everything else around. I can take care of this.

How common is it to install Vulkan in a custom directory? for CUDA and AMD ROCm/SMI, it's very common hence our dedicated configure options. If 99.99% of people are going to use distribution packages for Vulkan, we may just skip this and let the 0.01% use environment variable to help autoconf.

Given that you use some extensions in topology-vulkan.c, should we also query the PCI extension there instead of calling hwloc/vulkan.h which does its own query?

And we'll need some details (and better names) for VK_KHR_external_memory, VK_KHR_buffer_device_address, VK_KHR_timeline_semaphores. I assume you already use those?

Random questions:
* Is it possible to have multiple Vulkan devices under the same PCI device?
* Is there some sort of global device index that we could use instead of driver index + device index ? hwloc OpenCL devices are called OpenCLxdy where x is the plarform index and y is the device index in that platform. Vulkan looks similar but I don't like those names. So if there's anything better, I'd go for it.
* Is there any requirement to run a Xserver in the back? (GL had and it was annoying)
* Do you know how widely this code would work? I think I can mostly test it on Intel integrated GPU, and NVIDIA GPUs without a Xserver.

I don't think we'll have time to integrate this in the upcoming 2.14 but it's clearly meant to go the next stable release.

ct-clmsn:
> Your Cmake stuff splits the current code and then adds Vulkan stuff (and reindents some lines). I'll like to distinguish > those steps and just include Vulkan additions without moving everything else around. I can take care of this.

Apologies, I was teasing that stuff out for my personal readability. I can make a change this evening.

> How common is it to install Vulkan in a custom directory? for CUDA and AMD ROCm/SMI, it's very common hence our > dedicated configure options. If 99.99% of people are going to use distribution packages for Vulkan, we may just skip > this and let the 0.01% use environment variable to help autoconf.

Personally, I use the distribution packages. There two "main" implementations (to my knowledge): one you get from the distributions and one of the more prominent alternatives is [Lunar](https://vulkan.lunarg.com) which may install software in places that I'm unfamiliar with - an environment variable should work. If an environment variable is uniform for hwloc to manage this type of situation - that should do the trick.

> Given that you use some extensions in topology-vulkan.c, should we also query the PCI extension there instead of calling hwloc/vulkan.h which does its own query?

Sure thing.

> And we'll need some details (and better names) for VK_KHR_external_memory, VK_KHR_buffer_device_address, VK_KHR_timeline_semaphores. I assume you already use those?

I'm currently using all three for these reasons:

[VK_KHR_external_memory](https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_external_memory.html) indicates the device is capable of shared memory allocations with other APIs (like CUDA, OpenCL, Direct3D) or even separate OS processes without copying the data.

[VK_KHR_buffer_device_address](https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/buffer_device_address.adoc) indicates users access 64-bit GPU virtual address of a buffer (think "GPU pointers").

[VK_KHR_timeline_semaphores](https://www.khronos.org/blog/vulkan-timeline-semaphores) indicates the device is capable of counter-based semaphore for synchronization between the host and device.

I went down this road for Vulkan compute shaders and the shared memory capabilities are the most critical for HPC applications.

> Is it possible to have multiple Vulkan devices under the same PCI device?

Yes "[device groups](https://docs.vulkan.org/guide/latest/extensions/device_groups.html)". We could check for `VK_KHR_device_group` and `VK_KHR_device_group_creation`.

> Is there some sort of global device index that we could use instead of driver index + device index ? hwloc OpenCL > devices are called OpenCLxdy where x is the plarform index and y is the device index in that platform. Vulkan looks > similar but I don't like those names. So if there's anything better, I'd go for it.

Sure we could build a logical addressing (indexing?) scheme that's 1 dimensional over the (driver index, device index) pair. Think of it like a matrix stored column/row or row/column structured into a 1 dimensional array. It'd be a very simple indexing scheme that could be decoded on the fly. We'd just need to know the column and row lengths which can be collected at runtime.

> Is there any requirement to run a Xserver in the back? (GL had and it was annoying)

There is no need for Xserver. I've run this on headless systems (SiFive P550 and Raspberry PI 5 CM).

> Do you know how widely this code would work? I think I can mostly test it on Intel integrated GPU, and NVIDIA GPUs without a Xserver.

I've tested on MacOS w/MoltenVK, a SiFive P550, a Raspberry PI 5 CM, and WSL2 (Microsoft's Linux on Windows 2) all successfully.

bgoglin:
> Personally, I use the distribution packages. There two "main" implementations (to my knowledge): one you get from the distributions and one of the more prominent alternatives is [Lunar](https://vulkan.lunarg.com) which may install software in places that I'm unfamiliar with - an environment variable should work. If an environment variable is uniform for hwloc to manage this type of situation - that should do the trick.

If Lunar is common and a configure option will help its users, then let's keep the configure option then?

> [VK_KHR_external_memory](https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_external_memory.html) indicates the device is capable of shared memory allocations with other APIs (like CUDA, OpenCL, Direct3D) or even separate OS processes without copying the data.
> 
> [VK_KHR_buffer_device_address](https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/buffer_device_address.adoc) indicates users access 64-bit GPU virtual address of a buffer (think "GPU pointers").
> 
> [VK_KHR_timeline_semaphores](https://www.khronos.org/blog/vulkan-timeline-semaphores) indicates the device is capable of counter-based semaphore for synchronization between the host and device.

These are booleans then? Should we list all supported extensions? As separate keys in the array of info, or a single key with a comma-separated list of all extensions?

> > Is it possible to have multiple Vulkan devices under the same PCI device?
> 
> Yes "[device groups](https://docs.vulkan.org/guide/latest/extensions/device_groups.html)". We could check for `VK_KHR_device_group` and `VK_KHR_device_group_creation`.

I was asking this because your hwloc_vulkan_get_device_osdev() would need a way to distinguish between these multiple devices (we had the same issue for partitioned AMD devices). That said, maybe some of these functions in hwloc/vulkan.h are useless. They are basically a port of what we've been doing for all GPU API in the last 15 years, but maybe it's time to think about what's really useful.

> > Is there some sort of global device index that we could use instead of driver index + device index ? hwloc OpenCL > devices are called OpenCLxdy where x is the plarform index and y is the device index in that platform. Vulkan looks > similar but I don't like those names. So if there's anything better, I'd go for it.
> 
> Sure we could build a logical addressing (indexing?) scheme that's 1 dimensional over the (driver index, device index) pair. Think of it like a matrix stored column/row or row/column structured into a 1 dimensional array. It'd be a very simple indexing scheme that could be decoded on the fly. We'd just need to know the column and row lengths which can be collected at runtime.

I was rather asking for something that Vulkan already provides and users would already use.  If they used it, it would be useful to provide a function in hwloc/vulkan.h to convert it into a hwloc osdev. If they don't, they likely manipulate the device handle, hence we may just need a function to convert that handle into a osdev. Is there a way to get the platform handle back from the device handle? Or retrieve the device index or platform index from the handles?

> > Do you know how widely this code would work? I think I can mostly test it on Intel integrated GPU, and NVIDIA GPUs without a Xserver.
> 
> I've tested on MacOS w/MoltenVK, a SiFive P550, a Raspberry PI 5 CM, and WSL2 (Microsoft's Linux on Windows 2) all successfully.

Did you test on a multiple socket or multi NUMA platform where the PCI locality of a device may be different from the entire machine? That's what I tried in #776 but the extension always returned 0000:00:00.0 as the GPU PCI bus ID, which is clearly wrong (but I am not sure I was querying extensions properly). I'll try with your code on a machine with 2 server CPUs and 3 GPUs tomorrow.

ct-clmsn:
> If Lunar is common and a configure option will help its users, then let's keep the configure option then?

Sounds good.

> These are booleans then? Should we list all supported extensions? As separate keys in the array of info, or a single key with a comma-separated list of all extensions?

Correct - I'll need to take another look at the list of what's available to get a better sense of the suite of available extensions.

> I was asking this because your hwloc_vulkan_get_device_osdev() would need a way to distinguish between these multiple devices (we had the same issue for partitioned AMD devices). That said, maybe some of these functions in hwloc/vulkan.h are useless. They are basically a port of what we've been doing for all GPU API in the last 15 years, but maybe it's time to think about what's really useful.

Correct it was a good correlation point between between vulkan and what's been done previously. Solid refactor point.

> I was rather asking for something that Vulkan already provides and users would already use. If they used it, it would be useful to provide a function in hwloc/vulkan.h to convert it into a hwloc osdev. If they don't, they likely manipulate the device handle, hence we may just need a function to convert that handle into a osdev. Is there a way to get the platform handle back from the device handle? Or retrieve the device index or platform index from the handles?

I'll have to look into this approach.

> Did you test on a multiple socket or multi NUMA platform where the PCI locality of a device may be different from the entire machine? That's what I tried in https://github.com/open-mpi/hwloc/issues/776 but the extension always returned 0000:00:00.0 as the GPU PCI bus ID, which is clearly wrong (but I am not sure I was querying extensions properly). I'll try with your code on a machine with 2 server CPUs and 3 GPUs tomorrow.

Unfortunately, I don't have access to a system with that specific configuration. This is more of a proof of concept or a foundation to build upon, "minimum viable product".

bgoglin:
I just tested your code on a server with 3 NVIDIA GPUs.
PCI locality is wrong, similarly to what I got in my testing in #776. It gets PCI bus ID 0:0:2443223416.1 for all 3 devices which looks random. This uses VK_EXT_PCI_BUS_INFO_EXTENSION_NAME.

vulkaninfo gets the correct info from VkPhysicalDevicePCIBusInfoPropertiesEXT, so I guess both of us are not using the extension correctly but I couldn't find why when I looked at vulkaninfo source code.

If I disable that extension in your code to fallback to vkGetPhysicalDeviceProperties() instead, I get something wrong too. I don't understand you're code here. vkGetPhysicalDeviceProperties.deviceID is the device number, it's not related to a PCI bus ID.
```
$ lspci -d 10de: -n
41:00.0 0302: 10de:1eb8 (rev a1)
81:00.0 0302: 10de:1e78 (rev a1)
a1:00.0 0302: 10de:1eb8 (rev a1)
                   ^this is the deviceID
```

By the way, I get:
```
  CC       topology-vulkan.lo
../../hwloc/topology-vulkan.c:19:9: warning: "HWLOC_SHOWMSG_VULKAN" redefined
   19 | #define HWLOC_SHOWMSG_VULKAN 16
```
This line may just be removed.

bgoglin:
> CI fails because you need to add this:
> 
> ```
> --- i/include/hwloc/rename.h
> +++ w/include/hwloc/rename.h
> @@ -796,6 +796,7 @@ extern "C" {
>  #define hwloc_levelzero_component HWLOC_NAME(levelzero_component)
>  #define hwloc_nvml_component HWLOC_NAME(nvml_component)
>  #define hwloc_rsmi_component HWLOC_NAME(rsmi_component)
> +#define hwloc_vulkan_component HWLOC_NAME(vulkan_component)
>  #define hwloc_opencl_component HWLOC_NAME(opencl_component)
>  #define hwloc_pci_component HWLOC_NAME(pci_component)
>  
> ```

This shouldn't be needed anymore once you rebase on top of today's master (I disabled the need to rename internal symbols).

ctaylorTT:
could you run `vulkaninfo --json` and check for `VkPhysicalDevicePCIBusInfoPropertiesEXT`? there should be a value like `pciBus` in there with a hex bus address value. This should help validate against what lspci prints.

bgoglin:
Here's what I can find when looking for PCI:
```
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json:				"VkPhysicalDevicePCIBusInfoPropertiesEXT": {
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json-					"pciDomain": 0,
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json-					"pciBus": 129,
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json-					"pciDevice": 0,
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json-					"pciFunction": 0
VP_VULKANINFO_Quadro_RTX_8000_550_163_1_0.json-				},
--
VP_VULKANINFO_Tesla_T4_550_163_1_0.json:				"VkPhysicalDevicePCIBusInfoPropertiesEXT": {
VP_VULKANINFO_Tesla_T4_550_163_1_0.json-					"pciDomain": 0,
VP_VULKANINFO_Tesla_T4_550_163_1_0.json-					"pciBus": 65,
VP_VULKANINFO_Tesla_T4_550_163_1_0.json-					"pciDevice": 0,
VP_VULKANINFO_Tesla_T4_550_163_1_0.json-					"pciFunction": 0
VP_VULKANINFO_Tesla_T4_550_163_1_0.json-				},
```
The machine actually has two T4 GPUs but the json is only created for one of them (while both are shown in the normal output).
```
$ lspci -d 10de:
41:00.0 3D controller: NVIDIA Corporation TU104GL [Tesla T4] (rev a1)
81:00.0 3D controller: NVIDIA Corporation TU102GL [Quadro RTX 6000/8000] (rev a1)
a1:00.0 3D controller: NVIDIA Corporation TU104GL [Tesla T4] (rev a1)
```
PCI buses reported by vulkaninfo (129 and 65) are correct (decimal values for lspci 0x81 and 0x41)

PR #780, opened 2026-05-25
diff: report name appearing/disappearing as too complex

dxbjavid:
With two topologies where the same object has a name on one side and not on the other, `hwloc_diff_trees()` emits a `HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME` entry with either `oldvalue` or `newvalue` set to NULL. The XML export then dereferences NULL:

```
* thread #1, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: libsystem_platform.dylib`_platform_strlen + 4
    frame #1: hwloc__nolibxml_export_escape_string(src=0x0) at topology-xml-nolibxml.c:565
    frame #2: hwloc__nolibxml_export_new_prop(state, name="obj_attr_newvalue", value=0x0) at topology-xml-nolibxml.c:641
    frame #3: hwloc__xml_export_diff(parentstate, diff) at topology-xml.c:2917
    frame #4: hwloc_topology_diff_export_xmlbuffer(diff, ...)
```

The XML diff schema and `hwloc_apply_diff_one()` both require non-NULL old/new values, so a name appearing or disappearing cannot round-trip; report it as too_complex like the surrounding subtype/os_index/cpuset mismatches already do.

### DIFF
--- a/hwloc/diff.c
+++ b/hwloc/diff.c
@@ -156,8 +156,12 @@ hwloc_diff_trees(hwloc_topology_t topo1, hwloc_obj_t obj1,
 
 	/* gp_index don't have to be strictly identical */
 
-	if ((!obj1->name) != (!obj2->name)
-	    || (obj1->name && strcmp(obj1->name, obj2->name))) {
+	if ((!obj1->name) != (!obj2->name))
+		/* the NAME diff requires both old and new values; name appearing
+		 * or disappearing cannot be represented and won't round-trip
+		 * through XML import/export or apply */
+		goto out_too_complex;
+	if (obj1->name && strcmp(obj1->name, obj2->name)) {
                 err = hwloc_append_diff_obj_attr_string(topo1, obj1,
 						       HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME,
 						       NULL,


bgoglin:
I am tempted to convert this NULL from/to an empty string in a XML since an empty name is meaningless.

dxbjavid:
I thought about that too, but I’m slightly concerned it changes the meaning of the diff instead of only fixing the crash.

Right now NULL means the object has no name, while "" would mean it has a name that happens to be empty. The diff/apply code currently treats these as real values.

That is  why I leaned toward reporting it as 'too_complex', similar to the nearby subtype/os_index/cpuset cases, since the current XML schema doesn’t really represent a name appearing/disappearing cleanly.

But if the intended behavior is to treat empty string as equivalent to “no name” everywhere, then converting during XML export/import could also make sense.

bgoglin:
The libxml backend (widely used on Linux) exports an empty string for NULL, the crash doesn't occur there, so it's actually better to do the same for the nolibxml case.
Here's a patch that seems to work:
```
diff --git i/hwloc/topology-xml.c w/hwloc/topology-xml.c
index 34327a44f..1ed393492 100644
--- i/hwloc/topology-xml.c
+++ w/hwloc/topology-xml.c
@@ -1845,11 +1845,14 @@ hwloc__xml_import_diff_one(hwloc__xml_import_state_t state,
        break;
       case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_INFO:
        diff->obj_attr.diff.string.name = strdup(obj_attr_name_s);
-       /* FALLTHRU */
-      case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME:
        diff->obj_attr.diff.string.oldvalue = strdup(obj_attr_oldvalue_s);
        diff->obj_attr.diff.string.newvalue = strdup(obj_attr_newvalue_s);
        break;
+      case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME:
+        /* libxml export NULL strings as empty, and it empty name would be meaningless */
+       diff->obj_attr.diff.string.oldvalue = obj_attr_oldvalue_s ? strdup(obj_attr_oldvalue_s) : NULL;
+       diff->obj_attr.diff.string.newvalue = obj_attr_newvalue_s ? strdup(obj_attr_newvalue_s) : NULL;
+       break;
       }
 
       if (*firstdiffp)
@@ -2909,13 +2912,16 @@ hwloc__xml_export_diff(hwloc__xml_export_state_t parentstate, hwloc_topology_dif
        sprintf(tmp, "%llu", (unsigned long long) diff->obj_attr.diff.uint64.newvalue);
        state.new_prop(&state, "obj_attr_newvalue", tmp);
        break;
-      case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME:
       case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_INFO:
-       if (diff->obj_attr.diff.string.name)
-         state.new_prop(&state, "obj_attr_name", diff->obj_attr.diff.string.name);
+        state.new_prop(&state, "obj_attr_name", diff->obj_attr.diff.string.name);
        state.new_prop(&state, "obj_attr_oldvalue", diff->obj_attr.diff.string.oldvalue);
        state.new_prop(&state, "obj_attr_newvalue", diff->obj_attr.diff.string.newvalue);
        break;
+      case HWLOC_TOPOLOGY_DIFF_OBJ_ATTR_NAME:
+        /* name may be NULL and libxml exports NULL string as "" */
+       state.new_prop(&state, "obj_attr_oldvalue", diff->obj_attr.diff.string.oldvalue ? diff->obj_attr.diff.string.oldvalue : "");
+       state.new_prop(&state, "obj_attr_newvalue", diff->obj_attr.diff.string.newvalue ? diff->obj_attr.diff.string.newvalue : "");
+        break;
       }
 
       break;
```

dxbjavid:
I hadn’t realized the libxml backend was already exporting NULL strings as empty strings.

I tested the patch locally and it fixes the crash on the nolibxml path while keeping behavior consistent with the libxml backend. Preserving NULL again during import for NAME diffs also makes sense to avoid changing the in memory representation.

The patch looks good to me.

bgoglin:
Thanks for checking, I'll push this and it'll be in 2.14 (to be released soon). If you want to be credited better than as "dxbjavid", feel free to provide a real name :)

dxbjavid:
Thanks! Glad the report was useful.

You can credit me as Javid Khan.

PR #781, opened 2026-05-26
xml: free imported object leaked on invalid child tag

dxbjavid:
hwloc__xml_import_object() leaks the not-yet-inserted object on an unknown child tag because the subnode loop uses `goto error` instead of `goto error_with_object`.


bgoglin:
Thanks! Coverity should have found this one, surprising it didn't.

PR #782, opened 2026-05-26
xml: reject distances2 nbobjs whose square overflows unsigned

dxbjavid:
`nbobjs*nbobjs` at topology-xml.c:1236 is computed in 32-bit unsigned, so a `<distances2>` with nbobjs >= 65536 in untrusted XML wraps it to a tiny or zero malloc while the values matrix is still indexed over the full nbobjs. Reject nbobjs whose square overflows before allocating.

### DIFF
--- a/hwloc/topology-xml.c
+++ b/hwloc/topology-xml.c
@@ -1232,6 +1232,15 @@ hwloc__xml_import_distances(hwloc_topology_t topology,
     goto out;
   }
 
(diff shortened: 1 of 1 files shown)


bgoglin:
Thanks. Do you expect to have many other pull requests? I am planning to release 2.14.0rc1 soon. I could wait a couple more days to include some other fixes from you, but I won't wait for all of them if it lasts several weeks :)

PR #792, opened 2026-05-29
linux: reject nbnodes whose square overflows the distance matrix size

dxbjavid:
look_sysfsnode() sizes the NUMA distance matrix with nbnodes*nbnodes*sizeof(uint64_t), but nbnodes*nbnodes is evaluated as unsigned int and wraps before the widening to size_t. A crafted sysfs root (HWLOC_FSROOT) whose /sys/devices/system/node/online lists 65536 nodes makes the product 0, so malloc hands back a tiny buffer that hwloc_parse_nodes_distances() then overruns with nbnodes*nbnodes uint64 writes. Guard the square before allocating, like the other matrix-sizing paths.

### DIFF
--- a/hwloc/topology-linux.c
+++ b/hwloc/topology-linux.c
@@ -3424,6 +3424,13 @@ look_sysfsnode(struct hwloc_topology *topology,
   if (!indexes)
     return 0;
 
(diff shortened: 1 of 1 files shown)


bgoglin:
The CI slave died for an unrelated reason, we can ignore it.

PR #795, opened 2026-05-29
pci: null-terminate sysfs value buffer before parsing

dxbjavid:
The vendor/device/link sysfs reads in hwloc_look_pci fread up to sizeof(value) bytes into value[16] but never terminate it, then pass the buffer to strtoul/atof/atoi. A sysfs file that fills all 16 bytes makes those string parsers run past the buffer. Read one byte less and NUL-terminate, matching the rest of hwloc's sysfs parsing.


bgoglin:
I forgot this PR for some reason. Thanks.

dxbjavid:
Thank you.

PR #796, opened 2026-05-30
synthetic: set last level arity sentinel before processing indexes

dxbjavid:
hwloc_synthetic_process_indexes() scans data->level[] until it finds arity==0, but the deepest level's arity sentinel is only set after these calls run and the backend data comes from a plain malloc(), not zeroed. A synthetic description that interleaves indexes by a type which isn't matched before the last level (an absent type, reachable via hwloc_topology_set_synthetic and HWLOC_SYNTHETIC) then reads the uninitialized arity and can walk past data->level[]. Move the sentinel assignment ahead of the process_indexes loop.

### DIFF
--- a/hwloc/topology-synthetic.c
+++ b/hwloc/topology-synthetic.c
@@ -840,6 +840,12 @@ hwloc_backend_synthetic_init(struct hwloc_synthetic_backend_data_s *data,
     count++;
   }
 
+  /* the deepest level has no children, mark its arity sentinel before
+   * hwloc_synthetic_process_indexes() runs since the index interleaving
+   * loop scans data->level[] until it finds arity==0 (backend data is
+   * malloc'ed, not zeroed) */
+  data->level[count-1].arity = 0;
+
   /* set default attributes that depend on the depth/hierarchy of levels */
   for (i=0; i<count; i++) {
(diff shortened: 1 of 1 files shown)


bgoglin:
I am fine with the change but I don't understand the synthetic description case that you give. Could you give an example?

dxbjavid:
Sure. Something like `lstopo -i 'pack:2 core:2 pu:2(indexes=l2:core)'` (same string also works via hwloc_topology_set_synthetic() or HWLOC_SYNTHETIC).

The interleaving names L2, which isn't a level in that description, so the type scan in hwloc_synthetic_process_indexes() walks data->level[] looking for it and runs into the deepest level's arity. That entry is the sentinel that hadn't been zeroed yet, so before this patch it read uninitialized data and could keep scanning past level[] when the garbage was non-zero. With the sentinel set up front you just get the clean "Failed to find level for index interleaving loop type" error instead.

bgoglin:
Oh I didn't even remember we supported this syntax for interleaving. I need to fix an example in the doc and add a testcase for this in utils. Thanks!

bgoglin:
I added your example to the commit message and pushed it.

PR #797, opened 2026-05-31
tenstorrent hwloc support

ctaylorTT:
this is a draft pr to start upstreaming support for tenstorrent accelerators in hwloc.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Samuel Thibault              University of Bordeaux
 Jean-Yves VET                DDN
 Benjamin Worpitz
 Jeff Zhao                    Zhaoxin
+Christopher Taylor           Tenstorrent
 
 Affiliaion abbreviations:
 -------------------------
--- a/Makefile.am
+++ b/Makefile.am
@@ -81,7 +81,10 @@ endif HWLOC_BUILD_STANDALONE
 # Build the documenation and top-level README file
 #
 if HWLOC_BUILD_STANDALONE
-.PHONY: doc readme
+.PHONY: doc readme check-tenstorrent
 doc readme:
 	$(MAKE) -C doc
+# Delegate to tests/hwloc (do not use "make check TESTS=tenstorrent" at top level).
+check-tenstorrent:
+	$(MAKE) $(AM_MAKEFLAGS) -C tests/hwloc check-tenstorrent
 endif HWLOC_BUILD_STANDALONE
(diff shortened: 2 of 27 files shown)


ctaylorTT:
@bgoglin hello! wanted to post this draft PR to begin a dialogue about what you would want to see done before upstreaming. it offers a couple of support options:

* user installed tt-kmd (drivers) but tt-umd (user mode driver) and tt-metal (the sdk) are not available
* user installed drivers, user mode driver, and the sdk but doesn't want to include a yaml parser (tt-metal has some hardware description files available in yaml)
* user installed drivers, user mode driver, the sdk, and wants the yaml parser

bgoglin:
Hello,

I never heard of this accelerator :) This code works with the driver already in upstream Linux ?

Do we really need /dev/tenstorrent at configure time on the *build* node? Or could we build hwloc on the frontend node without tt hardware and then use in on compute node?

Are you sure you won't have dedicated library+headers in the future, which means we'll replace most of topology-torrent.c with calls to that library?

These info attributes shouldn't be needed since they are already in the PCI parent attribute:
```
<dt>PCIBusID=0000:04:00.0</dt>
<dt>PCIVendor=0x1e52</dt>
<dt>PCIDevice=0xfaca</dt>
<dt>PCISubsystemVendor=0x1e52</dt>
<dt>PCISubsystemDevice=0x0001 (Tenstorrent OS devices)</dt>
```
These may just be called Vendor and Model or something else without the GPU prefix. We don't remove the existing GPU prefix to avoid breaking existing users, but in your case I think it's better to use something generic:
```
<dt>GPUVendor=Tenstorrent</dt>
<dt>GPUModel=Tenstorrent Wormhole B0 (Tenstorrent OS devices)</dt>
```

Regarding tt-kmd, tt-umd, tt-metal, it's hard for me to decide. I am tempted to just support the normal case without bringing a dependency on yaml. However it's important for me to get a way to test the code without having the hardware, hence if tt-umd or tt-metal help there, I am interesting in hearing about it.

ctaylorTT:
@bgoglin thank you for the review! The code works with the current downstream driver. The tt-metal SDK has an OpenMPI dependency and with that the chain of dependencies with hwloc being one of them - which is motivating this effort. I'll unroll the yaml dependency along with the portions that integrate directly with tt-metal just to cut down on build dependencies. Will work on removing those PCI attributes and put attention on the Vendor/Model preference. As for testing, I'll be in touch so we can coordinate.

PR #798, opened 2026-06-01
linux: check strchr result when iterating PCI busids in find_osdev_parent

dxbjavid:
hwloc_linuxfs_find_osdev_parent() advances along the PCI busid chain of a sysfs device link with `tmp = strchr(tmp+4, ':')+9` (topology-linux.c:5747), assuming the domain is always 4 hex digits. With a sysfs root taken from a container or HWLOC_FSROOT, a link target like `.../pci0000:00/0:0:0.0` still matches the `%x:%x:%x.%x` sscanf but makes strchr return NULL, so the next iteration parses from `(char*)0x9`. Keep the result in a temporary and only advance when a ':' is actually found.


dxbjavid:

PR #799, opened 2026-06-02
linux: bound cxl uport busid parsing to link target length

dxbjavid:
annotate_cxl_dax reads `*(pcirootbus+11)` (and may write `*slash`) past the uportpath buffer when a crafted CXL uport symlink is shorter than the assumed busid layout.


dxbjavid:

PR #800, opened 2026-06-02
xml: avoid NULL parent deref when importing v2 bridge/pcidev as root

dxbjavid:
hwloc__xml_import_object() generates PCI localities for v2 topologies by reading obj->parent->cpuset, but never checks obj->parent for NULL. The loader doesn't force the root object to be a Machine, and Bridge/PCIDev are special types that skip the cpuset checks, so a crafted v2 document whose root is a Bridge with downstream PCI reaches this branch with a NULL parent and crashes at load (set_xml/set_xmlbuffer). Guard obj->parent like the checks just above.


dxbjavid:

PR #804, opened 2026-06-12
use unsigned char for ctype calls on possibly-signed input

dxbjavid:
hwloc_decode_from_base64() reads the encoded bytes from a signed char buffer and passes them straight to isspace(), so on platforms where char is signed any byte with the top bit set is sign-extended to a negative int. isspace() (and the other ctype functions) are only defined for EOF and values representable as unsigned char, so this is undefined and can index before the start of the ctype table on libc implementations that do not offset for negative indices. The bytes here are base64 object userdata coming from imported XML, so they are attacker-controlled. The same ctype-on-char pattern is in the hwloc_strncasecmp() fallback and in two of the command-line helpers, so I have cast the argument to unsigned char at each site to keep them consistent. No behaviour change for normal ASCII input.

### DIFF
--- a/include/private/misc.h
+++ b/include/private/misc.h
@@ -371,7 +371,7 @@ static __hwloc_inline int hwloc_strncasecmp(const char *s1, const char *s2, size
   return strncasecmp(s1, s2, n);
 #else
   while (n) {
-    char c1 = tolower(*s1), c2 = tolower(*s2);
+    char c1 = tolower((unsigned char)*s1), c2 = tolower((unsigned char)*s2);
     if (!c1 || !c2 || c1 != c2)
       return c1-c2;
     n--; s1++; s2++;
--- a/utils/hwloc/hwloc-calc.h
+++ b/utils/hwloc/hwloc-calc.h
@@ -350,7 +350,7 @@ hwloc_calc_parse_range(const char *_string,
   memcpy(string, _string, len);
   string[len] = '\0';
 
-  if (!isdigit(*string)) {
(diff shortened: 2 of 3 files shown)


bgoglin:
I'd rather not touch base64.c unless strictly necessary (e.g. build failure, etc). It's an external source file that I'd rather keep in sync with the original one. It's been working file for years for many other projects outside of hwloc.

bgoglin:
> I'd rather not touch base64.c unless strictly necessary (e.g. build failure, etc). It's an external source file that I'd rather keep in sync with the original one. It's been working file for years for many other projects outside of hwloc.

What's funny is that the latest base64.c (in OpenBSD, that's where we took it from) has some similar changes on lines that are very close but different from yours: https://github.com/libressl/openbsd/commit/7cbd1512a41f0043c3c5509a727f89428878eadb

I am going to import most changes from OpenBSD.

bgoglin:
Please rebase and see if your changes to base64.c are really important enough to make us diverge from upstream. Other changes are OK

dxbjavid:
Rebased and dropped the base64.c hunks. The OpenBSD import you pulled in already casts at each read site (ch = (unsigned char)*src++) which covers the isspace calls there, so no reason to diverge from upstream anymore. What's left is just the three one-line casts in hwloc_strncasecmp, hwloc_utils_parse_flags and hwloc_calc_parse_range. Build and the object userdata XML round-trip test still pass here.

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #805, opened 2026-06-13
hwloc: Fix heap buffer overflow when inserting synthetic NUMA level

saddamr3e:
Fix a heap buffer overflow in the synthetic topology parser when inserting an implicit NUMA level into a synthetic topology description that reaches the maximum supported depth.

## Root Cause

In `hwloc_backend_synthetic_init()`, the NUMA-level insertion path shifts existing levels upward using:

```c
memmove(&data->level[2],
        &data->level[1],
        count * sizeof(struct hwloc_synthetic_level_data_s));
```

When parsing reaches the maximum valid depth (`count == HWLOC_SYNTHETIC_MAX_DEPTH - 1`), this copies one element too many.

Only levels `[1 .. count-1]` need to be shifted because the machine root at index 0 remains in place. Copying `count` elements writes one structure past the end of the fixed-size `data->level[]` array.

## Fix

Change the move count from `count` to `count - 1` and document the invariant with an explanatory comment.

## Reproducer

A standalone reproducer has been added that constructs an all-explicit synthetic topology description with 125 `Group` levels followed by a `PU` level, causing the NUMA insertion path to execute with `count == 127`.

When built with AddressSanitizer:

* Before this patch: ASan reports a heap-buffer-overflow write in `hwloc_backend_synthetic_init()`.
* After this patch: the reproducer exits cleanly.

## Impact

This issue is reachable through attacker-controlled synthetic topology descriptions supplied via:

* `HWLOC_SYNTHETIC`
* `hwloc_topology_set_synthetic()`

and results in heap memory corruption during `hwloc_topology_load()`.

### DIFF
--- a/hwloc/topology-synthetic.c
+++ b/hwloc/topology-synthetic.c
@@ -826,13 +826,7 @@ hwloc_backend_synthetic_init(struct hwloc_synthetic_backend_data_s *data,
     /* insert a NUMA level below the automatic machine root */
     if (show_errors)
       fprintf(stderr, "hwloc/synthetic: Inserting a NUMA level with a single object at depth 1\n");
-    /* move existing levels by one.
-     * Only levels [1..count-1] need shifting to [2..count] (the level[0] machine
-     * root stays in place), i.e. count-1 elements. Using "count" here copied one
-     * extra (uninitialized) element and, when count reached its maximum of
-     * HWLOC_SYNTHETIC_MAX_DEPTH-1, wrote one struct past the end of the fixed-size
-     * data->level[] array (a heap buffer overflow reachable from an untrusted
-     * HWLOC_SYNTHETIC string / hwloc_topology_set_synthetic() input). */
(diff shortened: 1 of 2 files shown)


bgoglin:
Hello. Thanks. I reproduced by reducing MAX_DEPTH to 3, much easier. No need for the big comment in front of memmove, just put a single line like "count includes the root level which we are not memmov'ing".
I'll need a real name for thanking you, however the persons doing the PR and the commit look different?

saddamr3e:
Sorry about that
I forgot to switch my git config to my account before committing so the commit author did not match my PR account. I have fixed it now. I have also shortened the comment as you suggested.

bgoglin:
I squashed your commits into a single commit, removed the new test, added your PR explanation in the commit msg, and pushed. The squash used the first committer ID unfortunately, I forgot to check before pushing.
Thanks anyway.

PR #806, opened 2026-06-15
memtiers: free imported per-tier infos to avoid leak

dxbjavid:
The memory tier structures carry a set of info attributes that get populated when a topology is imported from XML, since a <memtier> element may contain <info> children. Those infos are heap allocated and handed over to the tier, but the teardown and restrict paths only release each tier's nodeset and never the infos, so the name/value strings and the array behind them are leaked on every load followed by destroy of such a topology. The duplication path has the matching gap, it copies the infos pointers from the old topology straight into the new one without duplicating them, which would turn into a double free once teardown started releasing them. I noticed it while comparing against the cpukinds code next door, which already frees and dups its infos in the same three places. The fix brings the memtier lifecycle in line: release the infos in destroy and in restrict, and deep copy them during dup, freeing the freshly duped nodeset if the infos copy fails. Checked on macOS with leaks, a topology whose XML has a memtier with an info child leaked 208 bytes before and is clean afterwards, and the test suite still passes.

### DIFF
--- a/hwloc/memattrs.c
+++ b/hwloc/memattrs.c
@@ -2098,6 +2098,12 @@ hwloc_internal_memtiers_dup(hwloc_topology_t new, hwloc_topology_t old)
       new->nr_memtiers = i;
       goto failed;
     }
+    if (hwloc__tma_dup_infos(tma, &tiers[i].infos, &old->memtiers[i].infos) < 0) {
+      assert(!tma || !tma->dontfree); /* this tma cannot fail to allocate */
+      hwloc_bitmap_free(tiers[i].nodeset);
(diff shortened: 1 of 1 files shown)


bgoglin:
Thanks, that's what happens when you decide late in the development of memtiers that we should have infos in case it's useful later, go back to add them in the first commit and forget to update the restrict() and dup() commits later.
By the way, I wondered in one of your commits why Coverity Scan wasn't finding all these bugs, it's been down for a while and it's not clear when/if it's coming back.
Thanks.

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #808, opened 2026-06-18
x86: bound otherids access to each PU level count in summarize

dxbjavid:
summarize() groups PUs by their extended-topology ids by reading infos[i].otherids[level] for every present PU using the level count of the last present PU (one). Each PU's otherids[] is only allocated to the number of levels its own cpuid 0x0b/0x1f leaf reported, and stays NULL when it reported none, so a PU with fewer levels than one, or a non-present index walked by the inner j loop, gets read past the end of its array or through a NULL pointer. This can show up on heterogeneous CPUs and is easy to trigger with a crafted cpuid dump replayed via HWLOC_CPUID_PATH, so guard the otherids access against a missing or shorter array at that level before dereferencing.


bgoglin:
Thanks

dxbjavid:
Thank you for the merge. Appreciated

PR #811, opened 2026-06-21
linux: stop node distance parsing at end of truncated file

dxbjavid:
hwloc_parse_nodes_distances reads each NUMA node's sysfs distance file into a fixed buffer and walks the space-separated values with strtoul, advancing tmp = next+1 after every number. If a node/distance file under a container or HWLOC_FSROOT sysfs holds fewer values than there are nodes and the last value runs right up to the buffer's terminating NUL, next ends up pointing at that NUL and next+1 steps one byte past the allocation, so the following strtoul reads out of bounds. We break out of the loop once next reaches the end of the string and let the existing incomplete-file check reject it; well-formed files are unaffected since they stop on found == nbnodes first.

### DIFF
--- a/hwloc/topology-linux.c
+++ b/hwloc/topology-linux.c
@@ -2757,6 +2757,9 @@ hwloc_parse_nodes_distances(unsigned nbnodes, unsigned *indexes, uint64_t *dista
       found++;
       if (found == nbnodes)
 	break;
+      if (!*next)
+	/* truncated file, no separator after the last value */
+	break;
       tmp = next+1;
     }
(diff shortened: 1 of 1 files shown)


bgoglin:
Hello. This looks ok, but tests failed for unrelated reasons (I changed some things on the CI slaves). Please rebase on top of git master so that make check should pass now.

dxbjavid:
Rebased on top of current master, make check should be happy now. Thanks for sorting out the CI.

dxbjavid:
Thank you for the merge.Appreciated

PR #812, opened 2026-06-23
darwin: update comment that _NSGetExecutablePath is not needed

kphuphanwoe:


bgoglin:
Our (old) Darwin CI slave has both __progname and getprogname, hence your code won't be used since it's at the end of the #if/#elif chain.
Did you see some improvement on some machine?

kphuphanwoe:
I completely overlooked that. My initial thought was just to ensure better coverage for Darwin, but you're right—it doesn't provide any actual improvement, it's just using a different Darwin-specific function.
Would you like me to close this PR, or should I make some changes based on your feedback?

bgoglin:
If you find some platforms where you can improve the situation, keep the PR. If not, maybe just modify the comment saying that _NSGetExecutablePath() doesn't seem need given that other alternative are supported.

kphuphanwoe:
It looks like no platforms would improve from this, so I'll remove this part and add a comment instead.

kphuphanwoe:
Rebased to current master and resolved conflicts

bgoglin:
Thanks

PR #815, opened 2026-06-26
x86: avoid signed shift overflow when extracting extended topology id

dxbjavid:
read_extended_topo derives a per-level id from the x2apic/extended-topology cpuid leaves, and the mask it builds shifts by apic_packageshift minus apic_shift. both shift amounts come straight from cpuid (eax & 0x1f), so when a dumped cpuid replayed through HWLOC_CPUID_PATH reports a package shift of 31, or a per-level shift larger than the package shift, the expression ends up evaluating 1<<31 or shifting by a wrapped-around count, which is undefined behaviour (ubsan flags both the left shift and the following subtraction). i've guarded the extraction so the id bits are only taken when the window is non-empty and switched the literal to unsigned, which leaves valid monotonic topologies unchanged (the window is empty there only when there are genuinely no bits to keep, where the old masked value was already 0).

### DIFF
--- a/hwloc/topology-x86.c
+++ b/hwloc/topology-x86.c
@@ -589,7 +589,16 @@ static void read_extended_topo(struct hwloc_x86_backend_data_s *data, struct pro
 	apic_nextshift = eax & 0x1f;
 	apic_type = (ecx & 0xff00) >> 8;
 	apic_id = edx;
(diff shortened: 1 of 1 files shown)


bgoglin:
I moved most of your big comment to the commit message and pushed it, thanks.

dxbjavid:
Thanks for taking the time to refine the comment and commit message. I appreciate the review and the merge.

PR #819, opened 2026-07-01
utils: cap oversized cell length when printing distance matrix

dxbjavid:
hwloc_utils_print_distance_matrix() formats every cell into a fixed 24-byte tmp buffer with snprintf, but snprintf returns the untruncated length, so a heterogeneous distances matrix whose OS device name is longer than 23 bytes (the name comes straight from imported XML, so it is attacker-controlled) leaves len larger than MATRIX_ITEM_SIZE_MAX and the following memcpy(buf + (MATRIX_ITEM_SIZE_MAX - len - 1), tmp, len+1) both reads past tmp and writes before the headers allocation, corrupting the heap when lstopo prints the distances of a topology loaded from an untrusted file. Cap len to MATRIX_ITEM_SIZE_MAX-1 after each snprintf so the offset stays in bounds and the cell is just displayed truncated.


dxbjavid:
Thank you for your time.

PR #820, opened 2026-07-03
utils: fix systemd-dbus-api buffer overflow on large cpusets

dxbjavid:
hwloc_utils_systemd_asprintf sizes its output buffer assuming the "ay 0x%04x" array-length prefix is always 9 bytes, but %04x is only a minimum width, so once the mask needs 0x10000 or more bytes the prefix grows to ten or more characters and the trailing bytes are written past the malloc'd buffer. It is reachable from hwloc-calc and lstopo when a cpuset or nodeset with a high bit (index 524280 or above) is converted with --cpuset-output-format/--nodeset-output-format systemd-dbus-api. I have budgeted the prefix for its widest form (up to eight hex digits for a 32bit byte count) so the allocation always has room; small masks format identically to before.


dxbjavid:
TY

PR #832, opened 2026-08-17
topology: fix dangling object returned by group dont_merge merge

dxbjavid:
When a Group with dont_merge set is inserted over an existing mergeable Group of the same cpuset, hwloc__insert_try_merge_group() calls hwloc_replace_linked_object(old, new), which moves the new object's contents into the old (still-linked) object and zeroes the new one, but then returns new instead of old. The caller hwloc__insert_object_by_cpuset() gets result==obj so its result!=obj cleanup is skipped and the moved-from struct leaks, and hwloc_topology_insert_group_object() hands the zeroed object (cpuset and attr NULL) back to the caller as if it were the inserted group. The four sibling branches that call hwloc_replace_linked_object() all return old, so this one should too.

dont_merge is settable from imported XML (hwloc__xml_import_object_attr), and this by-cpuset group merge is also reached via the public hwloc_topology_insert_group_object() API and distance-based grouping.

Reproduced with a synthetic topology: inserting a mergeable group over two packages, then inserting an equal group with dont_merge=1, returns an object with type 0 and cpuset/attr NULL before the change and the real in-tree group after it. Added a regression case to tests/hwloc/hwloc_groups.c that fails before and passes after; the whole tests/hwloc suite still passes.

### DIFF
--- a/hwloc/topology.c
+++ b/hwloc/topology.c
@@ -1592,7 +1592,7 @@ hwloc__insert_try_merge_group(hwloc_topology_t topology, hwloc_obj_t old, hwloc_
       /* keep the new one, it doesn't want to be merged */
       hwloc_replace_linked_object(old, new);
       topology->modified = 1;
-      return new;
+      return old;
 
     } else {
       if (old->attr->group.dont_merge)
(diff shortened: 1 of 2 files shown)


bgoglin:
Thanks. I am merging this and will add more asserts to the new test case.

dxbjavid:
TY!

PR #834, opened 2026-08-19
linux: grow midr cpukind infoarray to fit all six info fields

dxbjavid:
look_sysfscpukinds_by_midr_regs builds up to six info attributes per cpukind, LinuxCapacity plus the five MIDR sub-fields (implementer, architecture, variant, part, revision), but infoarray only has five slots. When two kinds differ in every MIDR field the common mask ends up 0, all five conditions add an entry, k reaches six and the sixth write lands one element past the array on the stack. The assert above only rules out the all-common case, so this isn't caught. It's reachable when loading from a crafted HWLOC_FSROOT whose hwloc-nofile-info reports aarch64 and whose per-cpu midr_el1 files differ in all five fields. Size the array for six entries.


dxbjavid:
Thanks!

